PR #1, opened 2016-10-18
Rework resource processing

ash2k:
Set template state to Ready when done
Set label on resources to point at template


ash2k:
@dadux

PR #2, opened 2017-01-08
Use client-go package

ash2k:
Rewrite most of Smith to use client-go package and provided utilities like cache.
There is a lot of work to do, but this already passes the integration test so I wanted to push it into the repo - maybe someone will want to do some hacking ;)

FYI @davecheney, @georgebarnett, @dadux, @benbarclay


pkg/app/app.go
```diff
@@ -2,170 +2,93 @@ package app
...
-	for _, tpr := range tprList.Items {
-		a.watchTpr(&tpr)
-	}
 
-	// 4. Watch for addition/removal of TPRs to start/stop watches
-	a.Watcher.Watch(smith.ThirdPartyResourceGroupVersion, smith.AllNamespaces, smith.ThirdPartyResourcePath, tprList.ResourceVersion, newTprEvent)
+	clients := dynamic.NewClientPool(a.RestConfig, nil, dynamic.LegacyAPIPathResolverFunc)
+
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
```
davecheney: ctx, cancel := context.WithCancel(ctx) ?

pkg/app/app.go
```diff
@@ -2,170 +2,93 @@ package app
...
+	defer cancel()
+
+	tp := processor.New(ctx, templateClient, clients, &StatusReadyChecker{})
+	defer tp.Join() // await termination
+	defer cancel()  // cancel ctx to signal done to processor (and everything else)
```
davecheney: this was already deferered on line 45

pkg/app/template_decoder.go
```diff
@@ -0,0 +1,30 @@
...
+	"encoding/json"
+
+	"github.com/atlassian/smith"
+
+	"k8s.io/client-go/pkg/runtime"
+	"k8s.io/client-go/pkg/watch"
+)
+
+type templateDecoder struct {
+	decoder *json.Decoder
```
davecheney: what about embedding an io.Closer here

type tempalteDecoder struct {
         io.Closer
         decoder *json.Decoder
}

Then you can delete the Close forwarding method below.

types.go
```diff
@@ -25,32 +31,36 @@ const (
...
+func (tl *TemplateList) GetObjectKind() unversioned.ObjectKind {
+	return &tl.TypeMeta
+}
+
+// GetListMeta is required to satisfy ListMetaAccessor interface.
+func (tl *TemplateList) GetListMeta() meta.List {
+	return &tl.Metadata
+}
+
+var _ runtime.Object = &TemplateList{}
```
davecheney: how about moving these to a test ? They don't need to be executed every time we run the program.

PR #13, opened 2017-03-09
Define annotations for output object name

ash2k:


docs/design/managing-resources.md
```diff
@@ -85,3 +85,94 @@ metadata:
...
+Please refer to their definition for definitions of `Tinst` and `K`.
+
+`smith.a.c/TprOutputNameModeField` can be applied to a TPR `T` to configure the way name of the output object `O`
+is constructed. Object `O` is a [Secret](https://kubernetes.io/docs/user-guide/secrets/) or a
+[ConfigMap](https://kubernetes.io/docs/user-guide/configmap/) which holds parameters that may be injected into
+object(s) `D` depending on `Tinst`. Object `O` is created/updated before `K` is created/updated.
+
+The following values of `smith.a.c/TprOutputNameModeField` are defined:
+- `Random` - `O`'s name is randomly generated on each `O` update.
+- `ContentsHash` - `O`'s name is a deterministic hash function of `O`'s contents. That means name changes iif contents change.
```
ojongerius: Maybe 

> `O`'s name is a deterministic hash function of it's contents. O's name changes if, and only if, it's contents change."

docs/design/managing-resources.md
```diff
@@ -85,3 +85,94 @@ metadata:
...
+- `Fixed` - `O`'s name is provided via a field on `Tinst`. Name of that field is defined by
```
ojongerius: Or:

> `O`'s name is provided via field `F` on `Tinst`. `F`'s name is defined by `smith.a.c/TprOutputNameSetterField` set on `T`.

docs/design/managing-resources.md
```diff
@@ -85,3 +85,94 @@ metadata:
...
+`smith.a.c/TprOutputNameSetterField` set on `T`.
+
+Each time `O`'s name changes `O` object is recreated, old object is deleted immediately and all dependent objects
+will have references to `O` updated.
```
ojongerius: I'm assuming we will want to create a new object, update references, and afterwards remove the old object? If so, maybe:

> Each time `O`'s name changes a copy `O` object is created, all dependent objects will have references updated `O` and the old object will be deleted.

PR #15, opened 2017-03-10
Switch to DeepCopy from conversion package.

ojongerius:


pkg/app/template_event_handler.go
```diff
@@ -27,12 +28,14 @@ func (h *templateEventHandler) OnDelete(obj interface{}) {
 
 func (h *templateEventHandler) handle(obj interface{}) {
 	in := obj.(*smith.Template)
-	out := &smith.Template{}
 
-	if err := smith.DeepCopy_Template(in, out); err != nil {
+	c := conversion.NewCloner()
```
ash2k: Wait, can we pass an instance of Scheme into this handler and use it's Copy method?

ash2k:
See #16

pkg/app/template_event_handler.go
ojongerius: We could, and done.

pkg/app/app.go
```diff
@@ -88,7 +88,10 @@ func (a *App) Run(ctx context.Context) error {
 		return errors.New("wait for Bundle Informer was cancelled")
 	}
 
-	sl := bundleStore{store: bundleInf.GetStore()}
+	sl := bundleStore{store: bundleInf.GetStore(),
```
ash2k: Put `store: ...` on a new line please like `scheme`

pkg/app/bundle_event_handler.go
```diff
@@ -4,10 +4,12 @@ import (
 	"log"
 
 	"github.com/atlassian/smith"
+	"k8s.io/apimachinery/pkg/runtime"
```
ash2k: Style: there should be a blank line between imports from same project and 3rd party imports.

pkg/app/bundle_event_handler.go
```diff
@@ -27,12 +29,13 @@ func (h *bundleEventHandler) OnDelete(obj interface{}) {
...
+	out, err := h.scheme.DeepCopy(in)
+	if err != nil {
 		log.Printf("[TEH] Failed to do deep copy of %#v: %v", in, err)
 		return
 	}
-	log.Printf("[TEH] Rebuilding %s/%s bundle because it was added/updated", out.Metadata.Namespace, out.Metadata.Name)
-	h.processor.Rebuild(out)
+
+	log.Printf("[TEH] Rebuilding %s/%s template because it was added/updated", out.(*smith.Bundle).Metadata.Namespace, out.(*smith.Bundle).Metadata.Name)
+	h.processor.Rebuild(out.(*smith.Bundle))
```
ash2k: Could extract `out.(*smith.Bundle)` into a variable and use it in all three places instead for better readability.

ojongerius:
Closing in favor of https://github.com/atlassian/smith/pull/21

PR #19, opened 2017-03-10
Fix for no kind "Status" is registered for version "v1"

ash2k:
/cc @nilebox


pkg/resources/client.go
```diff
@@ -35,6 +36,9 @@ func GetBundleTprClient(cfg *rest.Config) (*rest.RESTClient, *runtime.Scheme, er
 	if err := schemeBuilder.AddToScheme(scheme); err != nil {
 		return nil, nil, err
 	}
+	if err := api.AddToScheme(scheme); err != nil {
```
nilebox: @ash2k 👍 
client-go 👎

PR #20, opened 2017-03-10
Building dependency graphs with topological sort

nilebox:
**Not ready to merge**, need to add tests, more comments and make some code cleanup.
Also, no cycle detection.

- Building a graph of resources
- Splitting a graph into separate dependency subgraphs
- Topologically sorting each subgraph
- Independently visiting each subgraph to create resources with ready dependencies

The course of Algorithms and Data Structures by Robert Sedgewick :)

### DIFF
--- a/pkg/processor/graph/topological_sort.go
+++ b/pkg/processor/graph/topological_sort.go
@@ -1,0 +1,119 @@
+package graph
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/atlassian/smith"
+)
+
+// SortedData is a container for dependency graph and topological sort result
+type SortedData struct {
+	Graph          *Graph
+	SortedVertices []string
+}
+
+// TopologicalSort builds resource dependency graph and topologically sorts it
+func TopologicalSort(bundle *smith.Bundle) (*SortedData, error) {
+	graph := newGraph(len(bundle.Spec.Resources))
+
+	for _, res := range bundle.Spec.Resources {
+		graph.addVertex(res.Name)
+	}
+
+	for _, res := range bundle.Spec.Resources {
+		for _, d := range res.DependsOn {
+			graph.addEdge(res.Name, string(d))
+		}
+	}
+
+	sorted, err := graph.topologicalSort()
+	if err != nil {
+		return nil, err
+	}
+
+	graphData := SortedData{
+		Graph:          graph,
+		SortedVertices: sorted,
+	}
+
+	log.Printf("Sorted graph: %v", sorted)
+
+	return &graphData, nil
+}
+
+func (g *Graph) topologicalSort() ([]string, error) {
+	results := newOrderedSet()
+	for _, name := range g.orderedVertices {
+		err := g.visit(name, results, nil)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return results.items, nil
+}
+
+func (g *Graph) visit(name string, results *orderedset, visited *orderedset) error {
+	if visited == nil {
+		visited = newOrderedSet()
+	}
+
+	added := visited.add(name)
+	if !added {
+		index := visited.index(name)
+		cycle := append(visited.items[index:], name)
+		return fmt.Errorf("Cycle error: %s", strings.Join(cycle, " -> "))
+	}
+
+	n := g.Vertices[name]
+	for _, edge := range n.Edges() {
+		err := g.visit(edge, results, visited.clone())
+		if err != nil {
+			return err
+		}
+	}
+
+	results.add(name)
+	return nil
+}
+
+type orderedset struct {
+	indexes map[string]int
+	items   []string
+	length  int
+}
+
+func newOrderedSet() *orderedset {
+	return &orderedset{
+		indexes: make(map[string]int),
+		length:  0,
+	}
+}
+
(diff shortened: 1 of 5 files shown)


pkg/processor/worker.go
```diff
@@ -62,25 +63,46 @@ func (wrk *worker) rebuild(bundle *smith.Bundle) error {
...
+	log.Printf("Sorting resources")
+	// Split resources into dependency graphs and topologically sort them
+	graphs, err := graph.TopologicalSort(bundle)
+	if err != nil {
+		return err
+	}
+
+	// Visit each dependency subgraph
+	for _, names := range graphs {
+		// Visit resources of each dependency subgraph in topologically sorted order
```
nilebox: Iterate through each subgraph independently

pkg/processor/worker.go
```diff
@@ -62,25 +63,46 @@ func (wrk *worker) rebuild(bundle *smith.Bundle) error {
...
+			}
+			isReady, err := wrk.checkResource(bundle, resClone.(*smith.Resource))
+			if err != nil {
+				return err
+			}
+			if !isReady {
+				if err := wrk.setBundleState(bundle, smith.IN_PROGRESS); err != nil {
+					log.Printf("[WORKER] %v", err)
+				}
+				break
```
nilebox: If we found a not-ready-yet resource, break from the inner for loop and move to the next graph

pkg/processor/graph/disjoint_set.go
```diff
@@ -0,0 +1,79 @@
+package graph
+
+import "log"
+
+func (g *graph) disjointSubgraphs() []graph {
```
nilebox: Using Quick UnionFind algorithm to split the graph into disjoint subgraphs

pkg/processor/worker.go
ash2k: Cool, but `wrk.setBundleState()` should only be called after all sub graphs have been processed, right? :)

pkg/processor/graph/disjoint_set.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Build disjoint sets
+	disjointSets := make(map[string][]string)
+	for v := range g.vertices {
+		root := uf.find(v)
+		disjointSets[root] = append(disjointSets[root], v)
+	}
+
+	// Create corresponding subgraphs
+	subgraphs := make([]graph, uf.count, uf.count)
```
ash2k: If length and capacity is the same then `make([]graph, uf.count, uf.count)` is the same as `make([]graph, uf.count)`. Here and if few other places.

pkg/processor/graph/disjoint_set.go
```diff
@@ -0,0 +1,79 @@
...
+	i := 0
+	for _, ds := range disjointSets {
+		subgraphs[i] = *newGraph()
```
ash2k: `subgraphs[i]` can be extracted into a variable to avoid indexed access in loops below.

pkg/processor/graph/topological_sort.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	for _, res := range bundle.Spec.Resources {
+		for _, d := range res.DependsOn {
+			graph.addEdge(res.Name, string(d))
+		}
+	}
+
+	subgraphs := graph.disjointSubgraphs()
+	sorted := make([][]string, len(subgraphs), len(subgraphs))
```
ash2k: Can be simplified

pkg/processor/graph/topological_sort.go
```diff
@@ -0,0 +1,113 @@
...
+	for i := 0; i < len(subgraphs); i++ {
```
ash2k: `for i, subgraph := range subgraphs {`

pkg/processor/graph/types.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type vertex map[string]bool
+
+func (v vertex) addEdge(name string) {
+	v[name] = true
+}
+
+func (v vertex) edges() []string {
+	var keys []string
```
ash2k: `keys := make([]string, 0, len(v))`to pre-allocate slice

pkg/processor/worker.go
```diff
@@ -62,25 +63,46 @@ func (wrk *worker) rebuild(bundle *smith.Bundle) error {
 	}
 	log.Printf("[WORKER] Rebuilding the bundle %s/%s", wrk.namespace, wrk.bundleName)
 
+	// Build resource map by name
+	resourceMap := make(map[string]smith.Resource)
```
ash2k: Can pre-allocate map `make(map[string]smith.Resource, len(bundle.Spec.Resources))`

pkg/processor/graph/types.go
```diff
@@ -0,0 +1,52 @@
...
+	f.addEdge(to)
+	return nil
+}
+
+func (g *graph) containsVertex(name string) bool {
+	_, ok := g.vertices[name]
+	return ok
+}
+
+type vertex map[string]bool
```
ash2k: If `bool` is only used as a marker (is entry in the map?) then empty struct could be used to save a bit of space and make "used as marker" behaviour explicit: `type vertex map[string]struct{}`

PR #24, opened 2017-03-20
Ready check for Deployment

ash2k:


pkg/readychecker/built_in_types.go
```diff
@@ -0,0 +1,25 @@
+package readychecker
+
+import (
+	"github.com/atlassian/smith/pkg/resources"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	extensions_v1beta1 "k8s.io/client-go/pkg/apis/extensions/v1beta1"
```
nilebox: Isn't underscore against Golang style?

PR #25, opened 2017-03-21
Update client and use new cool stuff

ash2k:


pkg/readychecker/built_in_types.go
```diff
@@ -1,17 +1,18 @@
...
 )
 
+var converter unstructured_conversion.Converter = unstructured_conversion.NewConverter(false)
+
 // Works according to https://kubernetes.io/docs/user-guide/deployments/#the-status-of-a-deployment
 // and k8s.io/kubernetes/pkg/client/unversioned/conditions.go:120 DeploymentHasDesiredReplicas()
 func isDeploymentReady(obj *unstructured.Unstructured) (bool, error) {
 	var deployment extensions_v1beta1.Deployment
-	if err := resources.UnstructuredToType(obj, &deployment); err != nil {
+	if err := converter.FromUnstructured(obj.Object, &deployment); err != nil {
```
ash2k: Turns out there is now a built-in method for this.

pkg/processor/worker_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/atlassian/smith"
+	"github.com/atlassian/smith/pkg/resources"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestUpdateResourceEmptyMissingNilNoChanges(t *testing.T) {
```
ash2k: Need more tests to ensure spec is applied correctly in various situations.

pkg/processor/worker.go
```diff
@@ -301,63 +302,67 @@ func (wrk *worker) checkNeedsRebuild() bool {
...
-		valueClone, err := wrk.tp.scheme.DeepCopy(value)
+		valueClone, err := scheme.DeepCopy(value)
 		if err != nil {
 			return nil, err
 		}
 		updated.Object[field] = valueClone
 	}
 
-	if !reflect.DeepEqual(actualClone, actual) {
+	if !equality.Semantic.DeepEqual(updated, actualClone) {
```
ash2k: Use built-in comparator instead of bespoke normalizing functions.

pkg/processor/worker.go
```diff
@@ -301,63 +302,67 @@ func (wrk *worker) checkNeedsRebuild() bool {
...
+	delete(updated.Object, "status")
+
+	actClone, err := scheme.Copy(updated)
+	if err != nil {
+		return nil, err
+	}
+	actualClone := actClone.(*unstructured.Unstructured)
+
+	// This is to ensure those fields actually exist in underlying map whether they are nil or empty slices/map
+	actualClone.SetKind(actualClone.GetKind())
```
nilebox: can you elaborate what's the effect of this no-op (and below)?

ash2k: `Unstructured` is a `map[string]interface{}`. When field is missing in JSON it is missing in the map. `equality.Semantic.DeepEqual()` treats empt slices vs nil slices and empty maps vs nil maps as equal BUT missing fields vs existing fields are not equal. So those methods ensure corresponding fields are present in the map with whatever value there is.

nilebox: Then it should be `updated.GetKind()` instead of `actualClone.GetKind()`?

nilebox: `actual.GetKind()` actually

nilebox: `actual.GetKind()` actually

ash2k: At this point in code `actual` == `updated` == `actualClone`. I used the same object to make it more explicit that copying fields in not the purpose of the code.

PR #29, opened 2017-03-24
Use Store to Get resources

ash2k:
Should reduce number of api calls significantly.


pkg/processor/worker.go
```diff
@@ -157,58 +144,94 @@ func (wrk *worker) checkResource(bundle *smith.Bundle, res *smith.Resource) (isR
...
-		// TODO Maybe create a cache of Stores to avoid uselessly bombarding api server with requests?
-		// TODO How do we ensure we have informers for all object types so that we will get events once object is updated by its controller?
-		response, err = resClient.Get(name)
-		if err != nil {
-			if !errors.IsNotFound(err) {
+	if exists {
+		var ok bool
+		response, ok = obj.(*unstructured.Unstructured)
+		if !ok {
+			response = new(unstructured.Unstructured)
```
ash2k: No practical difference. Already changed it locally.

PR #37, opened 2017-04-02
Slices support

ash2k:


pkg/processor/spec_processor.go
```diff
@@ -62,7 +63,25 @@ func (sp *SpecProcessor) ProcessValue(value interface{}, path ...string) (interf
 		if err := sp.ProcessObject(v, path...); err != nil {
 			return nil, err
 		}
-		//TODO handle slices and slices of slices and ... inception. err, reflection
+	default:
+		// handle slices and slices of slices and ... inception. err, reflection
+		rv := reflect.ValueOf(value)
+		if rv.Kind() != reflect.Slice {
+			break
```
ash2k: This `default` case catches all types, including primitives which are ok. `SpecProcessor` should only be used on "JSON" objects (`map[string]interface{}`) so I don't want to handle crazy input here. If someone passes a struct, well, they'll get a struct back.

PR #52, opened 2017-05-03
Service Catalog support fixes and tests

ash2k:


integration_tests/main_test.go
```diff
@@ -38,16 +37,16 @@ func TestWorkflow(t *testing.T) {
 			Resources: bundleResources(t),
 		},
 	}
-	setupApp(t, bundle, false, testWorkflow)
+	setupApp(t, bundle, false, true, testWorkflow)
```
nilebox: two boolean flags are confusing, I would prefer passing a config struct with fields

integration_tests/service_catalog_test.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+		Spec: smith.BundleSpec{
+			Resources: []smith.Resource{
+				{
+					Name: smith.ResourceName(instance.Name),
+					Spec: toUnstructured(t, instance),
+				},
+				{
+					Name:      smith.ResourceName(binding.Name),
+					DependsOn: []smith.ResourceName{smith.ResourceName(instance.Name)},
```
nilebox: IIRC you can omit smith.ResourceName in the array declaration
```
DependsOn: []{smith.ResourceName(instance.Name)},
```

integration_tests/service_catalog_test.go
```diff
@@ -0,0 +1,89 @@
...
+					Spec:      toUnstructured(t, binding),
+				},
+			},
+		},
+	}
+	setupApp(t, bundle, true, true, testServiceCatalog)
+}
+
+func testServiceCatalog(t *testing.T, ctx context.Context, namespace string, bundle *smith.Bundle, config *rest.Config, clientset *kubernetes.Clientset,
```
nilebox: 10 input arguments + optional args... Is this a good practice in Go?

ash2k: It is not great, but this is just a test. I'll see if it would be better to put it all in a struct.

PR #60, opened 2017-05-24
Extract assertion func

ash2k:


integration_tests/utils_for_test.go
```diff
@@ -189,3 +189,22 @@ func createObject(t *testing.T, obj runtime.Object, namespace, resourcePath stri
...
+
+	assertCondition(t, bundleRes, smith.BundleReady, smith.ConditionTrue)
+	assertCondition(t, bundleRes, smith.BundleInProgress, smith.ConditionFalse)
+	assertCondition(t, bundleRes, smith.BundleError, smith.ConditionFalse)
+	assert.Equal(t, bundle.Spec, bundleRes.Spec, "%#v", bundleRes)
+
+	return bundleRes
+}
+
+func assertBundleTimeout(t *testing.T, ctx context.Context, store *resources.Store, namespace string, bundle *smith.Bundle) *smith.Bundle {
```
nilebox: Minor suggestion: usually `assert` comes with expectation in title.
For example, `assertTrue` ensures that the value equals `true`.
Something like `assertBundleWithTimeout` is better?

PR #61, opened 2017-05-24
Delete objects removed from Bundle

### LINKED ISSUE #56: Delete resources removed from a Bundle
If some resource or a tree of resources is removed from a `Bundle` (via an update of that `Bundle`), Smith should ensure removed objects are deleted.

Correct deletion order should be achieved and enforced by Owner References and finalizers. I.e. it should be enough to issue [synchronous delete](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/synchronous-garbage-collection.md) commands for all removed objects in any order and let GC+finalizers do the job.

See the list of blockers in #55.

The implementation should be able to recover if Smith fails in the middle of the deletion process - all orphaned objects (not in a `Bundle` and not marked for deletion and has `BundleName` label) can be found by inspecting their `bundle.smith.atlassian.com/BundleName` label, which has the name of the `Bundle` that should contain the object.

ash2k:
Fixes #56.

### DIFF
--- a/pkg/app/bundle_event_handler.go
+++ b/pkg/app/bundle_event_handler.go
@@ -5,6 +5,8 @@ import (
 	"log"
 
 	"github.com/atlassian/smith"
+
+	"k8s.io/client-go/tools/cache"
 )
 
 type bundleEventHandler struct {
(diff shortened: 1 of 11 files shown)


pkg/app/bundle_event_handler.go
```diff
@@ -22,22 +24,32 @@ func (h *bundleEventHandler) OnUpdate(oldObj, newObj interface{}) {
...
+	}
+	h.handle(obj, "deleted")
 }
 
-func (h *bundleEventHandler) handle(obj interface{}, addUpdate string) {
+func (h *bundleEventHandler) handle(obj interface{}, addUpdateDelete string) {
 	o, err := h.deepCopy(obj)
 	if err != nil {
-		bundle := o.(*smith.Bundle)
+		bundle := obj.(*smith.Bundle)
```
ash2k: Was using invalid object

PR #63, opened 2017-05-26
Consistent import aliases; use only versioned packages

ash2k:


pkg/readychecker/built_in_types.go
```diff
@@ -1,34 +1,31 @@
...
-	api_v0 "k8s.io/client-go/pkg/api"
-	apps_v0 "k8s.io/client-go/pkg/apis/apps"
-	appsv1beta1 "k8s.io/client-go/pkg/apis/apps/v1beta1"
-	extensions_v0 "k8s.io/client-go/pkg/apis/extensions"
-	extensions "k8s.io/client-go/pkg/apis/extensions/v1beta1"
-	settings_v0 "k8s.io/client-go/pkg/apis/settings"
+	api_v1 "k8s.io/client-go/pkg/api/v1"
+	apps_v1b1 "k8s.io/client-go/pkg/apis/apps/v1beta1"
+	ext_v1b1 "k8s.io/client-go/pkg/apis/extensions/v1beta1"
+	settings_v1a1 "k8s.io/client-go/pkg/apis/settings/v1alpha1"
```
ash2k: Here we use versioned packages now.

PR #72, opened 2017-06-09
Make Spec a runtime.Object

### LINKED ISSUE #71: Make Resource.Spec a runtime.Object
That way it is more convenient to use - any resource type can be used to populate the field.
`MarshalJSON()` and `UnmarshalJSON()` for `Resource` should be converting it to/from `unstructured.Unstructured`.

ash2k:
Fixes #71

### DIFF
--- a/integration_tests/tpr_attribute_test.go
+++ b/integration_tests/tpr_attribute_test.go
@@ -14,11 +14,22 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 func TestTprAttribute(t *testing.T) {
-	sleeper, sleeperU := bundleAttrResources(t)
+	sleeper := &tprattribute.Sleeper{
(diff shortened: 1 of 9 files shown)


integration_tests/tpr_attribute_test.go
```diff
@@ -14,11 +14,22 @@ import (
...
+		TypeMeta: meta_v1.TypeMeta{
+			Kind:       tprattribute.SleeperResourceKind,
+			APIVersion: tprattribute.SleeperResourceGroupVersion,
+		},
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name: "sleeper1",
+		},
+		Spec: tprattribute.SleeperSpec{
+			SleepFor:      1, // seconds,
+			WakeupMessage: "Hello, Infravators!",
```
nilebox: Change the text to "Hello world" or "World, wake up"? :)

integration_tests/utils_for_test.go
```diff
@@ -251,3 +257,7 @@ func assertBundleTimeout(t *testing.T, ctx context.Context, store *resources.Sto
 	defer cancel()
 	return assertBundle(t, ctxTimeout, store, namespace, bundle, resourceVersion)
 }
+
+func noCopy(src interface{}) (interface{}, error) {
```
nilebox: noop + comment that it implements deepCopy func

Issue #67, opened 2017-06-01
Handle adoption/orphaning properly


ash2k:
Currently Smith adopts an object (i.e. considers it belonging to a particular bundle) if an object with same name is defined in the bundle within the same namespace. It does not take into account whether that object has a "controller owner reference" to some other bundle or not. Behaviour should be as described in [ControllerRef proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/controller-ref.md). Only adopt (consider part of the bundle) an object if it does not have any "controller owner references". Adoption means adding a "controller owner reference" pointing to the owner bundle. If  an object that is mentioned in a bundle `A` has a controller (another bundle `B`) that owns it, then that bundle `A` should be marked with ERROR condition and IN PROGRESS and READY conditions should be set to false.

To add to the above - a decision needs to be made whether to implement adoption at all or not. What if the user wants to delete the bundle but keep some objects? Remove controller owner reference and "detach" an object from the bundle - should it adopt the object back or not? Or just point the controller owner reference to the object itself, delete the bundle and remove the controller reference?

ash2k:
So we'll go forward with "Do not adopt" approach. Remaining work to do:
- [ ] Detect which Bundle to rebuild based on controller reference, not on label. Still keep the label to aid filtering
- [ ] Need to have an index of object group+object kind+object name -> bundle in Bundle informer to ping specific bundle when an orphan object is deleted. This is needed because a bundle might want to create an object with same name so when it is deleted a bundle should be able to proceed. Similar to how we handle TPR updates.

ash2k:
This is done.

PR #75, opened 2017-06-13
Rework adoption

ash2k:
Updates #67

### DIFF
--- a/pkg/processor/worker.go
+++ b/pkg/processor/worker.go
@@ -205,7 +205,7 @@ func (wrk *worker) checkResource(bundle *smith.Bundle, res *smith.Resource, read
 	}
 
 	// 2. Create or update resource
-	resUpdated, retriable, err := wrk.createOrUpdate(spec)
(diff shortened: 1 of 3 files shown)


pkg/processor/worker.go
```diff
@@ -309,31 +309,41 @@ func (wrk *worker) createResource(resClient *dynamic.ResourceClient, spec *unstr
...
 		}
 	}
 
+	// Check that the object is not marked for deletion
+	if existingObj.GetDeletionTimestamp() != nil {
+		return nil, false, fmt.Errorf("object %v %q is marked for deletion", existingObj.GroupVersionKind(), existingObj.GetName())
+	}
+
+	// Check that this bundle owns the object
+	if !isOwner(existingObj, bundle) {
```
ash2k: So the proposed behaviour here is to NOT do adoption - i.e. if controller reference is missing or not pointing to the bundle it is not updated/added and an error is returned instead.
There is a small issue here - when an orphan that was taking up space and preventing bundle from progressing is deleted, nothing pokes the bundle so that it can retry to proceed. Need to think about it...

PR #76, opened 2017-06-13
Get controller Bundle based on owner references

ash2k:
Updates #67

### DIFF
--- a/pkg/app/resource_event_handler.go
+++ b/pkg/app/resource_event_handler.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/atlassian/smith"
+	"github.com/atlassian/smith/pkg/resources"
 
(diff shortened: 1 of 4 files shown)


pkg/app/resource_event_handler.go
```diff
@@ -26,14 +27,14 @@ func (h *resourceEventHandler) OnAdd(obj interface{}) {
 }
 
 func (h *resourceEventHandler) OnUpdate(oldObj, newObj interface{}) {
-	oldTmplName, oldNamespace := getBundleNameAndNamespace(oldObj)
+	oldBundleName, oldNamespace := getBundleNameAndNamespace(oldObj)
 
-	newTmplName, newNamespace := getBundleNameAndNamespace(oldObj)
+	newBundleName, newNamespace := getBundleNameAndNamespace(newObj)
```
ash2k: Test was failing because these two lines both used `oldObj`!

PR #78, opened 2017-06-14
Rebuild Bundle once conflicting object is deleted

ash2k:
Updates #67

### DIFF
--- a/pkg/app/resource_event_handler.go
+++ b/pkg/app/resource_event_handler.go
@@ -8,17 +8,22 @@ import (
 	"github.com/atlassian/smith/pkg/resources"
 
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


pkg/app/resource_event_handler.go
```diff
@@ -38,34 +43,46 @@ func (h *resourceEventHandler) OnUpdate(oldObj, newObj interface{}) {
...
 		}
-		meta, ok = tombstone.Obj.(meta_v1.Object)
+		metaObj, ok = tombstone.Obj.(meta_v1.Object)
 		if !ok {
 			log.Printf("[REH] Delete tombstone with unrecognized object type: %T", tombstone.Obj)
 			return
 		}
 	}
-	bundleName, namespace := getBundleNameAndNamespace(meta)
-	h.rebuildByName(namespace, bundleName, "deleted", obj)
```
ash2k: This line was using `obj` but it should be `meta`/`metaObj` because it is the variable that is updated when tombstones are handled (line 48/53).

PR #80, opened 2017-06-16
JsonPath support in references between objects

nilebox:


pkg/processor/spec_processor_test.go
```diff
@@ -11,6 +11,30 @@ import (
...
+		selfName:       "abc",
+		readyResources: readyResources(),
+		allowedResources: map[smith.ResourceName]struct{}{
+			"res1": {},
+		},
+	}
+	obj := map[string]interface{}{
+		"ref": map[string]interface{}{
+			// TODO: This JsonPath is ignored, fix the regex match expression
+			"slice": "str>$(res1#a.slice[?(@.label==\"label2\")].value)<str",
```
nilebox: @ash2k need to fix the regex in [spec_processor.go#L24](https://github.com/atlassian/smith/blob/master/pkg/processor/spec_processor.go#L24) (or replace regex with something else).

docs/design/managing-resources.md
```diff
@@ -48,7 +50,7 @@ spec:
         name: sleeper2
       spec:
         sleepFor: 4
-        wakeupMessage: "$(sleeper1/status/message)"
+        wakeupMessage: "$(sleeper1#status.message)"
```
nilebox: @ash2k does this format look fine to you? I wanted to decouple "resource name" reference from the JsonPath part.
`#` is like anchor on the GitHub documentation page section link: https://github.com/atlassian/smith#outputs

pkg/processor/spec_processor.go
```diff
@@ -126,7 +126,13 @@ func (sp *SpecProcessor) processMatch(selector string, primitivesOnly bool) (int
 	if _, allowed := sp.allowedResources[objName]; !allowed {
 		return nil, fmt.Errorf("references can only point at direct dependencies: %s", selector)
 	}
-	fieldValue := resources.GetNestedField(res.Object, names[1:]...)
+	// To avoid overcomplicated format of reference like this: $((res1#{$.a.string}))
+	// And have something like this instead: $((res1#a.string))
+	jsonPath := fmt.Sprintf("{$.%s}", parts[1])
```
nilebox: Initially I added support of full JsonPath, but it looked really ugly and unreadable: `$((res1#{$.a.string}))`.
Thus I added a shortened format support : `$((res1#a.string))`.
If we find some limitations (for example, some use cases where `$.` prefix is not appropriate), we can add a fallback logic (for example, don't add any prefix/suffix for the expressions starting with bracket `{`).

docs/design/managing-resources.md
```diff
@@ -8,12 +8,14 @@ Cross-object field references are a way for an object to get output field(s) of
...
 
-Syntax `"$({dependency1/fieldName})"` means that value of `fieldName` will be injected as a string instead
+Syntax `"$(dependency1#fieldName)"` means that value of `fieldName` will be injected as a string instead
 of the placeholder. In this case value must be a string, boolean or number.
 
-Syntax `"$(({dependency1/fieldName}))"` means that value of `fieldName` will be injected without quotes
+Syntax `"$((dependency1#fieldName))"` means that value of `fieldName` will be injected without quotes
 instead of the placeholder. In this case it can be a of any type including objects.
 
+The `fieldName` could be specified in JsonPath format (with `$.` prefix added by default), for example: `$(dependency1#status.conditions[?(@.type=="Ready")].status)`
```
nilebox: See the example of complex JsonPath reference:
`$(dependency1#status.conditions[?(@.type=="Ready")].status)`

pkg/processor/spec_processor.go
```diff
@@ -14,9 +14,9 @@ import (
 )
 
 const (
-	// . is a valid character in a name, cannot use as a separator.
```
nilebox: I don't think this concern was ever valid.  It's a Bundle resource name, not the Kubernetes object name.

pkg/resources/objects.go
```diff
@@ -34,3 +36,28 @@ func GetJsonPathString(obj map[string]interface{}, path string) (string, error)
...
+	j.AllowMissingKeys(true)
+	err := j.Parse(path)
+	if err != nil {
+		return nil, fmt.Errorf("JsonPath parse %s error: %v", path, err)
+	}
+	values, err := j.FindResults(obj)
+	if err != nil {
+		return nil, fmt.Errorf("JsonPath execute error: %v", err)
+	}
+	if len(values) == 0 {
```
ash2k: `switch len(values)` will be more readable?

pkg/resources/objects.go
```diff
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"fmt"
 
+	"errors"
```
ash2k: Format imports please

pkg/processor/spec_processor.go
ash2k: Well, I thought resource name in a bundle should follow the same naming guidelines but I'm fine with `#`.

pkg/resources/objects.go
nilebox: Not sure if it will, I prefer to run preconditions check with return before the actual processing logic.

nilebox:
@ash2k replace the old reference format with mustache-like:
`{{reference}}` and `{{{nakedReference}}}`
With old format JsonPath is impossible to implement without escape characters, which I would like to avoid (to keep it the same as plain JsonPath)

ash2k:
@nilebox this format is also fine. Something is wrong with tests though.

PR #86, opened 2017-06-22
Spec with defaults

### LINKED ISSUE #84: Fix regression introduced in #83
In #83 we added a temporary workaround for spec-vs-actual comparison. Now integration tests are failing. Obviously this need to be fixed asap.
Perhaps should address the issue properly. The only fix that comes to mind is applying defaults locally to the spec and then comparing with the "actual" object.

ash2k:
Fixes #84. There is more work to do to make it handle special fields properly.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -114,7 +114,7 @@ func (a *App) Run(ctx context.Context) error {
 	}
 
 	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "bundle")
-	cntrlr := controller.New(bundleInf, tprInf, bundleClient, bs, sc, rc, scheme.DeepCopy, rawStore, queue, a.Workers, a.ResyncPeriod, resourceInfs)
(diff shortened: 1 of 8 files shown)


pkg/app/app.go
```diff
@@ -114,7 +114,7 @@ func (a *App) Run(ctx context.Context) error {
 	}
 
 	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "bundle")
-	cntrlr := controller.New(bundleInf, tprInf, bundleClient, bs, sc, rc, scheme.DeepCopy, rawStore, queue, a.Workers, a.ResyncPeriod, resourceInfs)
+	cntrlr := controller.New(bundleInf, tprInf, bundleClient, bs, sc, rc, scheme, rawStore, queue, a.Workers, a.ResyncPeriod, resourceInfs)
```
nilebox: too many arguments for a method :/

types.go
```diff
@@ -176,6 +176,12 @@ func (bs *BundleStatus) ShortString() string {
 			buf.WriteString(cond.Reason)
 			buf.WriteByte('"')
 		}
+		if cond.Message != "" {
+			buf.WriteByte(' ')
```
nilebox: Isn't it better to write the result of `fmt.Sprintf()`? It would make it more readable

PR #121, opened 2017-08-08
Latest version of service catalog uses https

wryun:
This is a quick hack. May not want to merge...

### DIFF
--- a/cmd/smith/main.go
+++ b/cmd/smith/main.go
@@ -58,6 +58,9 @@ func runWithContext(ctx context.Context) error {
 	if *scUrl != "" {
 		scConfig := *config // shallow copy
 		scConfig.Host = *scUrl
+		scConfig.TLSClientConfig = rest.TLSClientConfig{
+			Insecure: true,
+		}
 		a.ServiceCatalogConfig = &scConfig
 	}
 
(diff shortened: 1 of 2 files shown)


wryun:
@ash2k @nilebox

cmd/smith/main.go
```diff
@@ -58,6 +58,9 @@ func runWithContext(ctx context.Context) error {
 	if *scUrl != "" {
 		scConfig := *config // shallow copy
 		scConfig.Host = *scUrl
+		scConfig.TLSClientConfig = rest.TLSClientConfig{
+			Insecure: true,
+		}
```
ash2k: I wonder what certificates it uses....

wryun: I got an error message about the cert being valid for some 10.* address (whereas I'm running smith outside kube targeting it at 192.168...).

Didn't bother investigating further, since the 'recommended' solution for service catalog is to skip-tls-verify:

https://github.com/kubernetes-incubator/service-catalog/blob/master/docs/walkthrough.md

(search for skip-tls-verify)

wryun: (maybe if you run in cluster it works?)

ash2k: I think I saw an issue about TLS and certificates with IP instead of a hostname in custom api servers somewhere....

ash2k:
I'm going to merge this - the whole thing will be removed once we figure out how to work via aggregator.

PR #122, opened 2017-08-09
Fix CRD add watch

ash2k:


pkg/controller/controller_crd_event_handler.go
```diff
@@ -25,29 +25,21 @@ type watchState struct {
...
 	*BundleController
-	mx       sync.Mutex            // protects the map
 	watchers map[string]watchState // CRD name -> state
 }
 
 func (h *crdEventHandler) OnAdd(obj interface{}) {
 	crd := obj.(*apiext_v1b1.CustomResourceDefinition)
-	if crd.Name == smith.BundleResourceName {
-		return
+	if h.ensureWatch(crd) {
```
wryun: Is this meant to address the 'reacting before API has it available' issue? If so, I don't understand - it seems like we're just erroring out more cleanly, but we'll still miss establishing the watch later on?

pkg/controller/controller_crd_event_handler.go
```diff
@@ -64,25 +56,35 @@ func (h *crdEventHandler) OnDelete(obj interface{}) {
...
+	}
+	if _, ok := h.watchers[crd.Name]; ok {
+		return true
+	}
+	if !resources.IsCrdConditionTrue(crd, apiext_v1b1.Established) {
+		log.Printf("[CRDEH] Not adding a watch for CRD %s because it hasn't been established", crd.Name)
+		return false
+	}
+	if !resources.IsCrdConditionTrue(crd, apiext_v1b1.NamesAccepted) {
+		log.Printf("[CRDEH] Not adding a watch for CRD %s because it's names haven't been accepted", crd.Name)
```
wryun: it's -> its

pkg/controller/controller_crd_event_handler.go
```diff
@@ -25,29 +25,21 @@ type watchState struct {
...
-		h.watch(crd)
-	}()
-	h.rebuildBundles(crd, "added")
 }
 
 func (h *crdEventHandler) OnUpdate(oldObj, newObj interface{}) {
 	newCrd := newObj.(*apiext_v1b1.CustomResourceDefinition)
-	if newCrd.Name == smith.BundleResourceName {
-		return
+	if h.ensureWatch(newCrd) {
```
wryun: ah, ok, we get another message OnUpdate do we? Makes sense. What's the update?

ash2k: Yes.
Add event is received and it only informs us that a CRD object has been added. It may not have been processed by the CRD controller yet - we can tell it by looking at the conditions. Then, once controller has processed the CRD, it updates it with proper conditions and we get an update event.

In case when CRD existed before Smith was started, add event may deliver an object that has the conditions set properly. So we need to check them in both OnAdd and OnUpdate. I've added comments.
Does it make sense?

PR #131, opened 2017-09-04
Use Bazel to build/test/release

ash2k:
Apart from goodness that Bazel provides, the motivation is to be able to use [code generation](https://github.com/kubernetes/code-generator) via special rules (this is a Bazel term) from Kubernetes so that we don't have to check in generated code. Namely client and deep copy code. This is the direction Kubernetes is going and we should follow. I'm talking both about not checking in generated code and using Bazel.

I think we will be able to use generators once we are on 1.8 machinery.


ash2k:
All `BUILD.bazel` files are autogenerated, however some of them have been modified - mostly added things. See comments.

.travis.yml
```diff
@@ -10,14 +10,37 @@ go:
...
 script:
-  - make build-all-race-ci
+  - make verify
+  - make update-bazel
   - make test-ci
-  - make coveralls
+  - make build-ci
+
+after_success:
+  - if [ "$TRAVIS_OS_NAME" == "linux" -a ! -z "$TRAVIS_TAG" ]; then
```
ash2k: This will build a docker image when building a tagged commit. Same approach as in gostatsd.

BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load("@io_bazel_rules_go//go:def.bzl", "gazelle", "go_library", "go_prefix")
+
+go_prefix("github.com/atlassian/smith")
+
+gazelle(
```
ash2k: gazelle is a tool that keeps lists of source files in `BUILD.bazel` files up to date. See https://github.com/bazelbuild/rules_go/#generating-build-files

WORKSPACE
```diff
@@ -0,0 +1,27 @@
...
+load("@io_bazel_rules_go//go:def.bzl", "go_repositories")
+load("@io_bazel_rules_docker//docker:docker.bzl", "docker_repositories", "docker_pull")
+
+go_repositories(
+    go_version = "1.9",
+)
+
+docker_repositories()
+
+docker_pull(
```
ash2k: Base layer for our Docker image(s)

cmd/smith/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+    visibility = ["//visibility:private"],
+    deps = [
+        "//cmd/smith/app:go_default_library",
+        "//pkg/client:go_default_library",
+        "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
+        "//vendor/k8s.io/client-go/rest:go_default_library",
+    ],
+)
+
+go_binary(
```
ash2k: Build binary

cmd/smith/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+    name = "smith",
+    importpath = "github.com/atlassian/smith/cmd/smith",
+    library = ":go_default_library",
+    visibility = ["//visibility:public"],
+)
+
+go_binary(
```
ash2k: Build binary with race detector

cmd/smith/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+    name = "smith-race",
+    gc_linkopts = ["-race"],
+    library = ":go_default_library",
+    visibility = ["//visibility:public"],
+)
+
+docker_build(
```
ash2k: Build docker image

cmd/smith/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+    name = "docker",
+    base = "@official_busybox//image",
+    cmd = ["/smith"],
+    files = [":smith"],
+)
+
+docker_push(
```
ash2k: Push image

it/BUILD.bazel
```diff
@@ -0,0 +1,62 @@
...
+        "resource_deletion_test.go",
+        "update_test.go",
+        "workflow_test.go",
+    ],
+    importpath = "github.com/atlassian/smith/it",
+    library = ":go_default_library",
+    tags = [
+        "external",
+        "manual",
+    ],
```
ash2k: Special tags https://docs.bazel.build/versions/master/test-encyclopedia.html#tag-conventions

PR #143, opened 2017-10-10
Update dependencies to Kubernetes 1.8 release

ash2k:
This is a preview of the PR, not ready for merge. Will be mergeable once Service Catalog releases a new version. This is to avoid depending on an unreleased version that is not available to deploy.


pkg/apis/smith/v1/doc.go
```diff
@@ -1,5 +1,3 @@
-// +k8s:deepcopy-gen=package,register
```
ash2k: @nilebox note how I removed package-wide deep copy generation and registration and instead marked specific types with `+k8s:deepcopy-gen=true`. That way I don't get the ugly `init()` and don't get the unnecessary deep copy methods on everything.

ash2k: Details here https://github.com/kubernetes/code-generator/blob/master/cmd/deepcopy-gen/main.go

ash2k: I mean, we should do the same in Service Catalog.

ash2k: Hm, in this case I actually don't see any difference (apart from omitted `init()`)...

PR #153, opened 2017-10-23
Pass namespace to crd event handler

amckague:
How should we test this?

### DIFF
--- a/pkg/controller/controller_crd_event_handler.go
+++ b/pkg/controller/controller_crd_event_handler.go
@@ -25,8 +25,7 @@ type watchState struct {
 type crdEventHandler struct {
 	ctx context.Context
 	*BundleController
-	watchers  map[string]watchState // CRD name -> state
-	namespace string
+	watchers map[string]watchState // CRD name -> state
 }
 
 // OnAdd handles just added CRDs and CRDs that existed before CRD informer was started.
--- a/pkg/controller/controller_crd_event_handler.go
+++ b/pkg/controller/controller_crd_event_handler.go
@@ -25,7 +25,8 @@ type watchState struct {
 type crdEventHandler struct {
 	ctx context.Context
 	*BundleController
-	watchers map[string]watchState // CRD name -> state
+	watchers  map[string]watchState // CRD name -> state
+	namespace string
 }
 
 // OnAdd handles just added CRDs and CRDs that existed before CRD informer was started.
@@ -91,7 +92,7 @@ func (h *crdEventHandler) ensureWatch(crd *apiext_v1b1.CustomResourceDefinition)
 		Kind:    crd.Spec.Names.Kind,
 	}
 	log.Printf("[CRDEH] Configuring watch for CRD %s", crd.Name)
-	res, err := h.smartClient.ForGVK(gvk, meta_v1.NamespaceNone)
+	res, err := h.smartClient.ForGVK(gvk, h.namespace)
 	if err != nil {
 		log.Printf("[CRDEH] Failed to setup informer for CRD %s: %v", crd.Name, err)
 		return false
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/ash2k/stager"
 	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
 	"github.com/atlassian/smith/pkg/cleanup"
 	clean_types "github.com/atlassian/smith/pkg/cleanup/types"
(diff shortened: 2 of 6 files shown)


pkg/controller/controller_crd_event_handler.go
```diff
@@ -25,7 +25,8 @@ type watchState struct {
 type crdEventHandler struct {
 	ctx context.Context
 	*BundleController
-	watchers map[string]watchState // CRD name -> state
+	watchers  map[string]watchState // CRD name -> state
+	namespace string
```
ash2k: This is not necessary - `crdEventHandler` struct embeds `*BundleController` which has the field.

ash2k:
> How should we test this?

We can test this by adding a testcase to the `controller_test`. Given a CRD and a namespace we should verify that a namespaced watch is established.

pkg/controller/controller_test.go
```diff
@@ -20,18 +22,19 @@ import (
...
-
-	"github.com/ash2k/stager"
 	scClientset "github.com/kubernetes-incubator/service-catalog/pkg/client/clientset_generated/clientset"
 	scFake "github.com/kubernetes-incubator/service-catalog/pkg/client/clientset_generated/clientset/fake"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	core_v1 "k8s.io/api/core/v1"
+	crv1 "k8s.io/apiextensions-apiserver/examples/client-go/apis/cr/v1"
 	apiext_v1b1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
ash2k: This import is here already - one line above :)

pkg/controller/controller_test.go
```diff
@@ -20,18 +22,19 @@ import (
 	"github.com/atlassian/smith/pkg/resources/apitypes"
 	"github.com/atlassian/smith/pkg/speccheck"
 	"github.com/atlassian/smith/pkg/store"
-
-	"github.com/ash2k/stager"
```
ash2k: This is the correct place for it - it is not in `github.com/atlassian/smith`

pkg/controller/controller_test.go
```diff
@@ -20,18 +22,19 @@ import (
...
 	scClientset "github.com/kubernetes-incubator/service-catalog/pkg/client/clientset_generated/clientset"
 	scFake "github.com/kubernetes-incubator/service-catalog/pkg/client/clientset_generated/clientset/fake"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	core_v1 "k8s.io/api/core/v1"
+	crv1 "k8s.io/apiextensions-apiserver/examples/client-go/apis/cr/v1"
```
ash2k: We have a CRD in Smith repo - could you change to use it instead? See `examples/sleeper/client.go` `SleeperCrd()` function. It is used in integration tests. Can get an instance of CRD and set the desired Status on it.

pkg/controller/controller_test.go
```diff
@@ -69,6 +73,7 @@ type testCase struct {
 func TestController(t *testing.T) {
 	t.Parallel()
 	tr := true
+	testNamespace := "test:namespace"
```
ash2k: Column is not an allowed symbol. Probably doesn't matter in a test.

pkg/controller/controller_test.go
```diff
@@ -97,6 +102,56 @@ func TestController(t *testing.T) {
 				},
 			},
 			expectedActions: sets.NewString("DELETE=/api/v1/namespaces/n1/configmaps/map1"),
+			namespace:       meta_v1.NamespaceNone,
```
ash2k: Could you change it to `meta_v1.NamespaceAll`? Both are empty strings but `meta_v1.NamespaceAll` means "(watch) all namespaces". `meta_v1.NamespaceNone` should be used as namespace for cluster-wide objects. I.e. meaning of the constant is different.

pkg/controller/controller_test.go
```diff
@@ -97,6 +102,56 @@ func TestController(t *testing.T) {
...
+					},
+					Status: apiextensionsv1beta1.CustomResourceDefinitionStatus{
+						Conditions: []apiextensionsv1beta1.CustomResourceDefinitionCondition{
+							{Type: apiextensionsv1beta1.Established, Status: apiextensionsv1beta1.ConditionTrue},
+							{Type: apiextensionsv1beta1.NamesAccepted, Status: apiextensionsv1beta1.ConditionTrue},
+						},
+					},
+				},
+			},
+			bundle: &smith_v1.Bundle{
```
ash2k: We should have a function that returns this to stop copying...

pkg/controller/controller_test.go
```diff
@@ -97,6 +102,56 @@ func TestController(t *testing.T) {
...
+				ObjectMeta: meta_v1.ObjectMeta{
+					Name:      "bundle1",
+					Namespace: "n1",
+					UID:       "uid123",
+				},
+			},
+			expectedActions: sets.NewString("GET=/apis/cr.client-go.k8s.io/v1/namespaces/" + testNamespace + "/examples"),
+			namespace:       testNamespace,
+			test: func(t *testing.T, ctx context.Context, bundleController *BundleController, testcase *testCase) {
+				subContext, _ := context.WithTimeout(ctx, 1000*time.Millisecond)
```
ash2k:
```go
subContext, subCancel := context.WithTimeout(...)
defer subCancel()
```

pkg/controller/controller_test.go
```diff
@@ -97,6 +102,56 @@ func TestController(t *testing.T) {
...
+				bundleController.Run(subContext)
+			},
+			testHandler: fakeActionHandler{
+				response: map[path]fakeResponse{
+					{
+						method: "GET",
+						watch:  true,
+						path:   "/apis/cr.client-go.k8s.io/v1/namespaces/" + testNamespace + "/examples",
+					}: {
+						statusCode: 200,
```
ash2k: `http.StatusOK`

PR #154, opened 2017-10-26
Plugin specification

ash2k:


docs/design/plugins.md
```diff
@@ -0,0 +1,164 @@
...
+      spec:
+        # need data from a-binding-secret but transformed
+        # e.g. only keys which start with "FOO_" when full names of the keys are not known in advance.
+```
+
+## Specification
+
+Smith plugins are [Go plugins](https://golang.org/pkg/plugin/). They are eagerly loaded at Smith startup to detect
+issues early. Names of plugins to load are passed via command line.
+Each plugin publishes a `Process(smith_v1.Resource, map[smith_v1.ResourceName]Dependency) (TransformResult, error)`
```
amckague: s/TransformResult/ProcessResult

docs/design/plugins.md
```diff
@@ -0,0 +1,164 @@
...
+// For reference:
+//
+//type Dependency struct {
+//	Spec    smith_v1.Resource
+//	Actual  runtime.Object
+//	Outputs []runtime.Object
+//}
+//
+//type ProcessResult struct {
+//	Object runtime.Object
```
amckague: I take it runtime.Object is happy to be [object] and smith will handle the difference between [] and object?

docs/design/plugins.md
```diff
@@ -0,0 +1,164 @@
...
+
+Smith plugins are [Go plugins](https://golang.org/pkg/plugin/). They are eagerly loaded at Smith startup to detect
+issues early. Names of plugins to load are passed via command line.
+Each plugin publishes a `Process(smith_v1.Resource, map[smith_v1.ResourceName]Dependency) (TransformResult, error)`
+function.
+
+When Smith comes across a resource with `type: plugin` and `pluginName: foobar` it invokes
+the plugin `foobar`. To do that it inspects all the dependencies of the resource specified in `dependsOn` attribute
+and fetches their outputs to include in the invocation along with the dependencies themselves. Smith needs to
+recognize resource group/version/kinds to be able to fetch the output (if there is one).
```
amckague: Does this extend to allowing smith bundle objects to depend on objects from outside the bundle or is this the preferred mechanism to allow that to happen?

docs/design/plugins.md
```diff
@@ -0,0 +1,164 @@
...
+    pluginName: filter
+    pluginSpec:
+      apiVersion: servicecatalog.k8s.io/v1beta1
+      kind: ServiceInstance
+      name: b
+      spec:
+        filterByPrefix: "FOO_" # only keys which start with "FOO_"          
+```
+
+When the plugin `filter` is invoked, it returns the following object:
```
nilebox: Would be nice for the sample plugin implementation above actually returning this result.

docs/design/plugins.md
```diff
@@ -0,0 +1,164 @@
...
+//}
+//
+//type ProcessResult struct {
+//	Object runtime.Object
+//}
+
+func Process(resource smith_v1.Resource, dependencies map[smith_v1.ResourceName]smith_plugin.Dependency)  (smith_plugin.ProcessResult, error) {
+	// Do the processing
+	return smith_plugin.ProcessResult{
+		//Object: object literal here
```
nilebox: Implement `filter` plugin there for a more intuitive example?

docs/design/plugins.md
ash2k: What is `[]` and `[object]`? I don't understand...

PR #161, opened 2017-11-01
Handle classname and planname the way servicecatalog does

amckague:
How do we test this?


nilebox:
@amckague to test this we need integration tests against actual running Service Catalog I believe.

PR #166, opened 2017-11-29
Avoid feature "race" and use buildozer to set race=on on tests

ash2k:
Ref https://github.com/bazelbuild/rules_go/issues/1068


ianthehat:
Seems good to me.
I am going to try to add a feature to default tests to `race="on"` soon, but this will still be safe even then.

PR #167, opened 2017-12-06
Initial implementation of Smith plugins

wryun:
This is a squashed commit of work from @ash2k, @nilebox, and @wryun


cmd/smith/app/app.go
```diff
@@ -174,5 +189,26 @@ func (a *App) controller(bundleInf, crdInf cache.SharedIndexInformer, bundleClie
 	}
 	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "bundle")
 
-	return controller.New(bundleInf, crdInf, bundleClient, bundleStore, sc, rc, cStore, specCheck, queue, a.Workers, a.ResyncPeriod, resourceInfs, a.Namespace)
+	return controller.New(bundleInf, crdInf, bundleClient, bundleStore, sc, rc, cStore, specCheck, queue, a.Workers, a.ResyncPeriod, resourceInfs, a.Namespace, plugins, scheme)
```
wryun: This is starting to look insane... Mikhail assures me that someone will refactor it ;)

docs/design/plugins.md
```diff
@@ -126,7 +126,10 @@ spec:
       metadata:
         name: a
       spec:
-        foo: bar
+        clusterServiceClassExternalName: user-provided-service
+        clusterServicePlanExternalName: default
+        parameters:
+          foo: bar
```
yiyangc91: Not releated to this PR, but this should look like

```
parameters:
  credentials:
    foo: bar
```

For user-provided-service.

pkg/controller/controller_sync_task.go
```diff
@@ -156,7 +167,78 @@ func (st *syncTask) evalSpec(res *smith_v1.Resource) (*unstructured.Unstructured
...
+		if err != nil {
+			return nil, err
+		}
+		if err = unstructured_conversion.DefaultConverter.FromUnstructured(unstructuredActual.Object, actual); err != nil {
+			return nil, err
+		}
+		dependency.Actual = actual
+		switch obj := actual.(type) {
+		case *sc_v1b1.ServiceBinding:
+			secret, exists, err := st.store.Get(core_v1.SchemeGroupVersion.WithKind("Secret"), obj.Namespace, obj.Spec.SecretName)
```
nilebox: extract this block into a separate method?
also probably would be nice to have a single file with one such method per object type.

pkg/controller/controller_sync_task.go
```diff
@@ -156,7 +167,78 @@ func (st *syncTask) evalSpec(res *smith_v1.Resource) (*unstructured.Unstructured
...
+	// TODO validate pluginSpec GVK/name against result.Object.Spec
+	// and validate type of object against schema if possible
+
+	log.Printf("Plugin %s result: %+v", res.PluginName, result.Object)
+	return util.RuntimeToUnstructured(result.Object)
+}
+
+// evalNormalSpec evaluates the regular resource specification and returns the result.
+func (st *syncTask) evalNormalSpec(res *smith_v1.Resource) (*unstructured.Unstructured, error) {
+	// 0. Convert Spec to Unstructured
```
nilebox: nit: you count steps from 1 in `evalSpec` and from 0 here.
make consistent, or remove ordinary numbers completely.

cmd/smith/app/app.go
```diff
@@ -14,6 +16,7 @@ import (
 	smithClient_v1 "github.com/atlassian/smith/pkg/client/clientset_generated/clientset/typed/smith/v1"
 	"github.com/atlassian/smith/pkg/client/smart"
 	"github.com/atlassian/smith/pkg/controller"
+	smithPlugin "github.com/atlassian/smith/pkg/plugin"
```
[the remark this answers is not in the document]
ash2k: Using both underscores and camel case is not idiomatic I think :(
But Idea suggests camel case when it sees underscores (snake case). I don't care which case, but we need to be consistent. However I don't think this is something that is blocking this pr. But should be fixed eventually (soon).

pkg/apis/smith/v1/types.go
```diff
@@ -193,38 +195,48 @@ type Resource struct {
...
+	Kind       string `json:"kind"`
+	Name       string `json:"name"`
 }
 
 func (r *Resource) UnmarshalJSON(data []byte) error {
-	type resource struct {
-		Name      ResourceName              `json:"name"`
-		DependsOn []ResourceName            `json:"dependsOn"`
-		Spec      unstructured.Unstructured `json:"spec"`
+	var res struct {
```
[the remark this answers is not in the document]
ash2k: Because `Resource` has a `runtime.Object` field (`Spec`) and there is no way to unmarshal into it. So we use a temporary struct with `Spec unstructured.Unstructured`

pkg/controller/controller_sync_task.go
ash2k: This is a good idea

nilebox: But you didn't make this change before merging. Sigh...

ash2k: I did?

nilebox: @ash2k apologies, all good. I looked at the wrong revision I guess.

Issue #165, opened 2017-11-28
Consider using `go_image` to reduce boilerplate.


mattmoor:
See here: https://github.com/atlassian/smith/blob/63624b47d77f48cae9379528ead25d3fea00e9b5/cmd/smith/BUILD.bazel#L20-L32

This can use [this](https://github.com/bazelbuild/rules_docker#go_image):
```python
  go_image( 
     name = "smith", 
     embed = [":go_default_library"], 
     importpath = "github.com/atlassian/smith/cmd/smith", 
     visibility = ["//visibility:public"], 
 ) 
```

(the goal is to make containerizing as simple as `s/lang_binary/lang_image/`)

You might also TAL at `rules_k8s`, feel free to reach out if you have any problems.

ash2k:
Hey, thanks for suggestion. I new about these extra rules but I haven't tried them. I have just tried and here is what I found:
- it is cool that I don't need to maintain distroless base layer sha
- Gazelle still generates `go_binary` rule and if `go_image` has `name=smith` then they conflict.
- If I change `name=smith_container` on `go_image` to avoid the conflict then it works but now I have two rules - just like before :)
- I need an ability to run the binary. I usually build it and then run it (I don't use `bazel run` for this). This becomes `bazel-bin/cmd/smith/smith_container.binary` vs `bazel-bin/cmd/smith/smith`. It's a bit uglier and, which is more important, more magical - it is not obvious what is that `.binary` and there is no way to know it exists without looking at the rule implementation.

So given the above I think it is unlikely I'll use it unless there is a way to improve the user experience somehow.

mattmoor:
Yeah, I have a bug open against Gazelle.  This came up [here](https://github.com/kubernetes/test-infra/pull/4929#pullrequestreview-80041244) so I may augment the surface of language rules that don't benefit from layering to also accept the binary target directly.

(to elaborate a bit) For Go there is less benefit than with languages like Python and Java where we can intelligently factor layers (you can see some commentary on this [here](https://github.com/bazelbuild/rules_docker/issues/232) and [here](https://github.com/bazelbuild/rules_k8s/blob/master/examples/hellogrpc/py/server/BUILD#L24-L28)).  Eventually I'd like to make this more intelligent and less manual.

I'll open an FR on `rules_docker` to track this for now.

> it is cool that I don't need to maintain distroless base layer sha

[...]

PR #168, opened 2017-12-07
Use go_image rule to remove boilerplate

ash2k:
Ref #165. @mattmoor, hey this is cool! I like to have logically distinct things to be separate. Easier to understand it that way. Too much magic inside of a single rule is confusing.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -6,22 +6,14 @@ git_repository(
 git_repository(
     name = "io_bazel_rules_docker",
     remote = "https://github.com/bazelbuild/rules_docker.git",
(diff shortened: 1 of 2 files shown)


mattmoor:
@ash2k Ack, this is (to some extent) because I thought it was easier this way to achieve fine layering (in other languages).  We might be able to leverage this patten and still get it with Aspects (I hear it cures cancer!), but I haven't found time to play with them yet.

Let us know if you have any issues.  I (and others) lurk on the Bazel channel on the Kubernetes Slack if you ever want to chat a bit more synchronously.

PR #169, opened 2017-12-07
Prepare for plugin shim code

ghost:
This changes the Plugin code to allow a smith loaded "Plugin that handles Smith Plugins" to handle any sort of "Smith Plugin".

The main changes:

- Build the plugin and smith code using go build
- Dumb Shim plugin for now
- Add "IsSupported" as a function that a plugin may implement, to indicate what "smith plugins" it can handle. I guess the precedence is determined by plugin order specified on the command line.

I feel like the term "plugin" is a bit overloaded here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,15 @@
 METALINTER_CONCURRENCY ?= 4
 ALL_GO_FILES=$$(find . -type f -name '*.go' -not -path "./vendor/*" -not -path "./build/*" -not -path './pkg/client/clientset_generated/*' -not -name 'zz_generated.*')
-GOVERSION := 1.9.2
+
+# Go Build flags
 GP := /gopath
 GOPATH ?= "$$HOME/go"
+GOVERSION := 1.9.2
+ARCH ?= darwin
+BINARY_NAME := smith
+MAIN_PKG := github.com/atlassian/smith/cmd/smith
+PLUGIN_SHIM_NAME := shim
+PLUGIN_SHIM_PKG := github.com/atlassian/smith/plugins/shim
 
 setup: setup-ci
 	go get -u golang.org/x/tools/cmd/goimports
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -146,4 +153,30 @@ docker-export:
 release: update-bazel
 	bazel run --cpu=k8 //cmd/smith:push-docker
 
+.PHONY: docker-pull-go
```
yiyangc91: Probably want to fix up PHONY

ash2k:
Why do we need `IsSupported()`? Plugin is a single plugin with a name. If it does multiple things it should either be split into multiple plugins OR what exactly needs to be done should be provided to the plugin via `PluginSpec`. This complicates the code IMO but does not provide any new functionality - it is already possible to do what it does. WDYT?

pkg/plugin/types.go
```diff
@@ -11,10 +11,18 @@ import (
...
 )
 
-type Func func(resource smith_v1.Resource, dependencies map[smith_v1.ResourceName]Dependency) (ProcessResult, error)
+type Process func(resource smith_v1.Resource, dependencies map[smith_v1.ResourceName]Dependency) (ProcessResult, error)
+type IsSupported func(plugin string) (bool, error)
+
+type Plugin struct {
+	Name        string
+	Process     Process
+	IsSupported IsSupported
```
nilebox: So a single plugin can support multiple "Bundle plugin names"?
e.g. a single shim plugin can proxy calls to all "actual" plugin processes?
This looks fine, just not sure if using the "plugin" term in both context is correct there, it's a bit confusing.

The function can probably be called something like `Implements(...)`

nilebox: `Implements` is a bad name too I think :)

plugins/shim/plugin.go
```diff
@@ -0,0 +1,36 @@
...
+	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
+	smith_plugin "github.com/atlassian/smith/pkg/plugin"
+	sc_v1b1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Process(resource smith_v1.Resource, dependencies map[smith_v1.ResourceName]smith_plugin.Dependency) (smith_plugin.ProcessResult, error) {
+	// This is dumb, and just spits out a random service instance
+	name := resource.PluginSpec.Name
+	svcInstance := &sc_v1b1.ServiceInstance{
```
nilebox: So this is the place where we'll do the gRPC call or the REST call, or fork/pipe magic (whatever we agree to), right?

nilebox: if we go the RPC path, I think we should get rid of Go plugins completely, and just stick with the RPC mechanism we agree to. Reasoning for that:
- Go plugins are not mature in general
- It's an unnecessary layer of abtraction, since we unlikely to have other Go plugin implementations in the near future
- with Go plugins it will be hard to init plugin with config (list of supported plugins/endpoints, for example), and cache config between the calls (if we need it)
- it will make Smith build easier, and plugin builds will be a completely separate process. All that we'll need is a config for Smith, which can be either passed as command line parameters, or through ConfigMap stored in Kubernetes.

nilebox (commented):
The code is fine, but we might reconsider the way we support plugins in Smith (see comments).

cmd/smith/app/app.go
```diff
@@ -192,23 +189,49 @@ func (a *App) controller(bundleInf, crdInf cache.SharedIndexInformer, bundleClie
...
-		f, ok := symbol.(func(resource smith_v1.Resource, dependencies map[smith_v1.ResourceName]smith_plugin.Dependency) (smith_plugin.ProcessResult, error))
+
+		supportedFunc, ok := supportedSymbol.(func(plugin string) (bool, error))
 		if !ok {
-			return nil, errors.Errorf("loaded symbol from plugin %q does not have the right signature", p)
+			return nil, errors.Errorf("loaded IsSupported from plugin %q does not have the right signature", p)
 		}
-		plugs[p] = f
+		if supportedFunc == nil {
+			return nil, errors.New("OMG?????????")
```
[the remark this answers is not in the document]
yiyangc91: Oops. Debugging..

yiyangc91:
Okay, closing this for now until we decide what to do.

PR #171, opened 2017-12-07
Reworked plugin support

ash2k:


pkg/apis/smith/v1/types.go
```diff
@@ -187,6 +187,9 @@ func (bs *BundleStatus) ShortString() string {
 // ResourceName is a reference to another Resource in the same bundle.
 type ResourceName string
 
+// PluginName is a name of a plugin to be invoked.
```
wryun: what's the advantage of this? (curiosity)

cmd/smith/main.go
```diff
@@ -41,63 +28,9 @@ func runWithContext(ctx context.Context) error {
...
-
-	if *plugins != "" {
-		a.Plugins = strings.Split(*plugins, ",")
-	}
-
-	if pprofAddr != nil && *pprofAddr != "" {
-		go func() {
-			log.Fatalf("pprof server failed: %v", http.ListenAndServe(*pprofAddr, nil))
-		}()
+	if err := a.ParseFlags(flag.CommandLine, os.Args[1:]); err != nil {
```
wryun: I'd probably push as something that automatically happens on Run (and give CommandLine/Args). Just in the interest of simplicity of interface... I don't see any advantage to forcing the plugin author to call ParseFlags before calling run.

In a similar vein, why not just have an 'a.Run' command which sets up 'a' with a default workers of 2? And if you want a different number of workers, have it passed as a flag?

pkg/apis/smith/v1/types.go
ash2k: Type safety

cmd/smith/main.go
ash2k: 1. Originally `app` package did not depend on flags or configuration files or anything like that. It is/was agnostic of where the configuration comes from. This helps a lot when you want to reuse the code or set up the whole app in an integration test. Imagine setting up an integration test with `.Run()` that wants you to pass a FlagSet into it. Ugly. I was actually thinking to put flag parsing into a separate package and keeping `app` clean.
2. Workers should be a command line flag, yes.

PR #172, opened 2017-12-08
Add spec, and simplify plugin function sig

wryun:
This makes it clearer that it's Smith's job to worry about whether the plugin is spec compliant. I'm not sure about this now, but I'm leaving the PR here for discussion... you might have a plugin that wants to react to the specified type in the spec (i.e. pass PluginSpec instead of Spec), or even to something like the type annotations on the object.

I'm also still confused about why we want PluginSpec separate from Spec. Isn't everything simpler if we just have a single 'Spec' object, and just treat it separately depending on the type? What's the advantage of having a separate PluginSpec?

(I've targeted a branch for now so we can see the diff, but I don't want to merge this into the branch)


pkg/controller/controller_sync_task.go
```diff
@@ -183,17 +183,31 @@ func (st *syncTask) evalPluginSpec(res *smith_v1.Resource) (*unstructured.Unstru
...
 	})
 	if err != nil {
 		return nil, err
 	}
-	// TODO validate pluginSpec GVK/name against result.Object.Spec
-	// and validate type of object against schema if possible
 
-	log.Printf("Plugin %q result: %+v", res.PluginName, result.Object)
-	return util.RuntimeToUnstructured(result.Object)
+	// Make sure plugin is returning us something that obeys the PluginSpec.
```
wryun: I think we want this regardless.

ash2k:
Addressed in another pr.

PR #175, opened 2017-12-13
Make resourcespec properly have nil fields

wryun:
Otherwise we have non-nil type info, and != nil checks fail.

Also removed unnecessary logging

### DIFF
--- a/pkg/apis/smith/v1/types.go
+++ b/pkg/apis/smith/v1/types.go
@@ -227,7 +227,12 @@ func (rs *ResourceSpec) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
-	rs.Object = res.Object
+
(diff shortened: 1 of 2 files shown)


pkg/apis/smith/v1/types.go
```diff
@@ -227,8 +227,13 @@ func (rs *ResourceSpec) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
-	rs.Object = res.Object
-	rs.Plugin = res.Plugin
+	// If we do this copy, we get nils with type info (i.e. != nil)
+	if res.Object != nil {
+		rs.Object = res.Object
+	}
+	if res.Plugin != nil {
```
ash2k: This is unnecessary - `rs.Plugin` is not an interface.

PR #176, opened 2017-12-13
Use a closer to working plugin example

wryun:


docs/design/plugins.md
```diff
@@ -45,14 +48,11 @@ spec:
...
-        apiVersion: servicecatalog.k8s.io/v1beta1
-        kind: ServiceInstance
-        metadata:
-          name: b
+      plugin:
+        name: filter
+        objectName: b
         spec:
-          # need data from a-binding-secret but transformed
-          # e.g. only keys which start with "FOO_" when full names of the keys are not known in advance.
```
ash2k: This is an example of a situation where something like plugins is needed =)

docs/design/plugins.md
```diff
@@ -45,13 +48,12 @@ spec:
...
-        kind: ServiceInstance
-        metadata:
-          name: b
+      plugin:
+        name: filter
+        objectName: b
         spec:
           # need data from a-binding-secret but transformed
+          filterByPrefix: "FOO_" # only keys which start with "FOO_"
           # e.g. only keys which start with "FOO_" when full names of the keys are not known in advance.
```
ash2k: Once again, the point of this example is to illustrate the need for a mechanism to do runtime transformation of data. That is why it only has comments in the `spec` section. Does it make sense? I guess if you are trying to fix this it does not. How can we make this a better example?

docs/design/plugins.md
```diff
@@ -16,26 +16,29 @@ metadata:
...
   - name: a-binding
     dependsOn:
     - a
     spec:
       object:
-        apiVersion: servicecatalog.k8s.io/v1beta1
-        kind: ServiceBinding
         metadata:
           name: a-binding
+        apiVersion: servicecatalog.k8s.io/v1beta1
```
nilebox: nit: the canonical order (in all Kubernetes examples) is to have apiVersion and kind first, followed by metadata.
e.g. https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment

docs/design/plugins.md
wryun: I mean, see name of 'PR'? (closer to working example). I find it easier to understand what's going on with an actual field here. It's just an addition to your 'e.g.' (or rather, exactly what was here before?). Should it be after the 'e.g.' comment line?

wryun: ok, so, you want it to be an object - should I add the 'filter plugin' afterwards? It feels very strange to have the subsequent lines.

PR #179, opened 2017-12-18
Copy the service catalog UserInfo into cleaned up spec

ghost:


pkg/cleanup/types/built_in.go
```diff
@@ -113,6 +113,9 @@ func scServiceInstanceCleanup(spec, actual *unstructured.Unstructured) (*unstruc
 	instanceSpec.ObjectMeta.Finalizers = instanceActual.ObjectMeta.Finalizers
 
 	instanceSpec.Spec.ExternalID = instanceActual.Spec.ExternalID
+	if instanceActual.Spec.UserInfo != nil {
+		instanceSpec.Spec.UserInfo = instanceActual.Spec.UserInfo.DeepCopy()
+	}
```
yiyangc91: Anund mentioned that because we only care about the parameters section, we could probably copy a larger chunk of things/code could be simpler.

ash2k (approved):

pkg/cleanup/types/built_in.go
```diff
@@ -75,6 +75,9 @@ func scServiceBindingCleanup(spec, actual *unstructured.Unstructured) (*unstruct
 	}
 
 	sbSpec.Spec.ExternalID = sbActual.Spec.ExternalID
+	if sbActual.Spec.UserInfo != nil {
+		sbSpec.Spec.UserInfo = sbActual.Spec.UserInfo.DeepCopy()
```
nilebox: nice catch

pkg/cleanup/types/built_in.go
nilebox: yeah I can imagine making a deep copy of the entire spec and just overwriting parameters + parametersFrom

Issue #77, opened 2017-06-13
Only delete objects removed from Bundle once Bundle is ready


ash2k:
Otherwise objects might be deleted too early.

PR #180, opened 2017-12-19
Per-resource status

### LINKED ISSUE #99: Report status for each resource
To improve visibility/reporting and aid troubleshooting Smith should have status of each resource as part Bundle status.

### LINKED ISSUE #138: Ready checker does not handle failures for Service Catalog types

ash2k:
Fixes #99.
Fixes #138.


pkg/controller/bundle_sync_task.go
```diff
@@ -249,6 +298,46 @@ func (st *bundleSyncTask) isBundleReady() bool {
 	return true
 }
 
+// updateResourceCondition updates passed condition by fetching information from an existing resource condition if present.
+// Sets LastTransitionTime to now if the status has changed.
+// Returns true if resource condition in the bundle does not match and needs to be updated.
+func updateResourceCondition(b *smith_v1.Bundle, resName smith_v1.ResourceName, condition *smith_v1.ResourceCondition) bool {
```
ash2k: This is a copy of a function that updates bundle conditions but for resource conditions.

pkg/controller/resource_sync_task.go
```diff
@@ -327,7 +342,7 @@ func (st *resourceSyncTask) createResource(resClient dynamic.ResourceInterface,
 	if api_errors.IsAlreadyExists(err) {
 		// We let the next processKey() iteration, triggered by someone else creating the resource, to finish the work.
 		err = api_errors.NewConflict(schema.GroupResource{Group: gvk.Group, Resource: gvk.Kind}, spec.GetName(), err)
-		return nil, false, errors.Wrapf(err, "object %q found, but not in Store yet (will re-process)", spec.GetName())
+		return nil, false, errors.Wrap(err, "object found, but not in Store yet (will re-process)")
```
ash2k: Resource name is added to the error string up the call stack.

pkg/readychecker/types/built_in.go
```diff
@@ -63,7 +73,15 @@ func isScServiceInstanceReady(obj *unstructured.Unstructured) (isReady, retriabl
...
-	return readyCond != nil && readyCond.Status == sc_v1b1.ConditionTrue, false, nil
+	if readyCond != nil && readyCond.Status == sc_v1b1.ConditionTrue {
+		return true, false, nil
+	}
+	failedCond := getServiceInstanceCondition(&instance, sc_v1b1.ServiceInstanceConditionFailed)
+	if failedCond != nil && failedCond.Status == sc_v1b1.ConditionTrue {
+		return false, false, errors.Errorf("%s: %s", failedCond.Reason, failedCond.Message)
+	}
+	// TODO support "unknown" and "in progress"
+	return false, false, nil
```
ash2k: These changes fix #138. Now failure is recognized properly.

pkg/controller/bundle_sync_task.go
```diff
@@ -70,19 +71,24 @@ func (st *bundleSyncTask) process() (retriableError bool, e error) {
...
+		_, resErr := resInfo.fetchError()
+		blockedOnError = blockedOnError || resErr != nil
+		log.Printf("[WORKER][%s/%s] Resource %q, ready: %t, error: %v", st.bundle.Namespace, st.bundle.Name, resName, resInfo.isReady(), resErr)
 		st.processedResources[resName.(smith_v1.ResourceName)] = &resInfo
 	}
-	// Delete objects which were removed from the bundle
-	retriable, err := st.deleteRemovedResources()
-	if err != nil {
-		return retriable, err
+	if !blockedOnError {
```
wryun: I'd be tempted to just have a method which traverses processResources here rather than cache the state. Should be cheap? Or am I misunderstanding something?
---
Also, I'm trying to understand the exact condition under which objects are removed from the bundle. Shouldn't this object removal only take place once the bundle has completely finished? At the moment, it looks like it deletes after the initial pass, when some resources are still blocked/in progress? If this is like CF, this is not a good idea (for example, the old compute could still be up referring to a particular resource...)

pkg/apis/smith/v1/types.go
```diff
@@ -150,7 +172,8 @@ type BundleCondition struct {
 // +k8s:deepcopy-gen=true
 type BundleStatus struct {
 	// Represents the latest available observations of a Bundle's current state.
-	Conditions []BundleCondition `json:"conditions,omitempty"`
+	Conditions     []BundleCondition `json:"conditions,omitempty"`
+	ResourceStatus []ResourceStatus  `json:"resourceStatuses,omitempty"`
```
wryun: Hmm... I guess this is some style thing that this is ResourceStatus rather than ResourceStatuses. But still, confusing.

pkg/apis/smith/v1/types.go
```diff
@@ -258,3 +291,36 @@ func (rs *ResourceSpec) IntoTyped(obj runtime.Object) error {
 	}
 	return fmt.Errorf("cannot convert %T into typed object %T", rs.Object, obj)
 }
+
+// +k8s:deepcopy-gen=true
+type ResourceStatus struct {
+	Name       ResourceName        `json:"name"`
+	Conditions []ResourceCondition `json:"conditions,omitempty"`
```
wryun: But wait! This is Conditions[], not Condition[]!

wryun: (I would prefer all collections are plural - see above)

pkg/controller/bundle_sync_task.go
wryun: This should like it updatesResourceConditions, but it just checks. shouldUpdateResourceCondition? hasResourceConditionChanged?

pkg/controller/bundle_sync_task.go
ash2k: Why you think a method would be better?

> Shouldn't this object removal only take place once the bundle has completely finished?
This is bug #77, I know about this. I don't want to fix it without a test and this PR is too big already.

pkg/apis/smith/v1/types.go
ash2k: Yes, I was following style of some code from Kubernetes.

pkg/apis/smith/v1/types.go
ash2k: Hm, you are right. We can change it to be consistent.

pkg/apis/smith/v1/types.go
```diff
@@ -183,6 +206,16 @@ func (bs *BundleStatus) ShortString() string {
 	return buf.String()
 }
 
+func (bs *BundleStatus) GetResourceStatus(resName ResourceName) (int, *ResourceStatus) {
+	for i := range bs.ResourceStatus {
+		resStatus := &bs.ResourceStatus[i]
```
wryun: ok, so I assume you're doing this to extract a pointer to a slice element. But isn't this a bit weird, in that now you're assuming that slice will never be GC'd/mutated? (confused about golang, again)

wryun: (also, repeating what I said earlier about pointer returns being nasty :P)

pkg/controller/bundle_sync_task.go
wryun: I just don't like caching when it's not necessary (possibility of inconsistent state, etc.). When you move the handling of object removal out of here, there will be nothing so close to this traversal anyway.

pkg/apis/smith/v1/types.go
ash2k: So, as discussed, there is no obvious much better way to do this. I'll keep it a pointer for now. I've also changed the method on the Bundle to behave the same way.

pkg/controller/bundle_sync_task.go
ash2k: This flag is also passed to `resourceSyncTask` which needs it. I.e. it is needed during processing, not only after it. I don't see why using a method in this case is better in any way (I would probably agree about iteration at the end otherwise).

PR #181, opened 2017-12-19
Enable repository cache and update Gazelle

ash2k:


.bazelrc
```diff
@@ -1,3 +1,5 @@
+common --experimental_repository_cache=/tmp/bazel_experimental_repository_cache
```
ash2k: No more Go 1.9.2.tar.gz re-downloads on rules_go version bump!

PR #190, opened 2017-12-28
Initial schema for Bundle CRD

### LINKED ISSUE #139: Hot loop on invalid resource format
If a resource in a bundle is misspecified, smith hot loops.
```
E1004 04:09:38.667737       1 streamwatcher.go:109] Unable to decode an event from the watch stream: unable to decode watch event: Object 'Kind' is missing in '.....'
```

### LINKED ISSUE #141: Define OpenAPI schema Bundle CRD
This is necessary to prevent storage of invalid bundles.
https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#validation

ash2k:
Fixes #141.
Fixes #139.


cmd/smith/app/app.go
```diff
@@ -207,10 +210,35 @@ func NewFromFlags(flagset *flag.FlagSet, arguments []string) (*App, error) {
 	flagset.IntVar(&a.Workers, "workers", 2, "Number of workers that handle events from informers")
 	flagset.StringVar(&a.Namespace, "namespace", meta_v1.NamespaceAll, "Namespace to use. All namespaces are used if empty string or omitted")
 	pprofAddr := flag.String("pprof-address", "", "Address for pprof to listen on")
+	printBundleSchema := flag.String("print-bundle-schema", "", "Print Bundle schema and exit (specify format: json or yaml)")
```
nilebox: what do you need this application mode for? debugging?

docs/deployment/0-crd.yaml
```diff
@@ -11,3 +11,147 @@ spec:
...
+            apiVersion:
+              $ref: '#/definitions/apiVersion'
+            kind:
+              $ref: '#/definitions/kind'
+          required:
+          - kind
+          - apiVersion
+          type: object
+      properties:
+        spec:
```
ash2k: This schema only validates the `spec` of the `Bundle`. Common metadata is validated by kubernetes itself.

cmd/smith/app/app.go
```diff
@@ -207,10 +210,35 @@ func NewFromFlags(flagset *flag.FlagSet, arguments []string) (*App, error) {
...
+		data, err := yaml.Marshal(resources.BundleCrd().Spec.Validation.OpenAPIV3Schema)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to marshal Bundle schema")
+		}
+		_, err = os.Stdout.Write(data)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed write schema YAML to stdout")
+		}
+		return nil, context.Canceled
+	case "":
```
nilebox: do you actually need a special `case ""` there? isn't `default` enough?
could be some code style thing I guess.

nilebox: if the flag is not set, you should skip printing a bundle schema and run the application I suppose?
as far as I see currently the default value is `""` (empty string), and you will exit with error int that case

nilebox: ah nvm, Go `switch` works differently from other languages (https://tour.golang.org/flowcontrol/9): 
> Go only runs the selected case, not all the cases that follow

can you add some comment there like
```go
case "":
  // flag not set, do nothing
```

cmd/smith/app/app.go
ash2k: To get the schema in json or yaml? Because I don't want to write it by hand.

cmd/smith/app/app.go
ash2k: Added a comment

cmd/smith/app/app.go
nilebox: should be a separate binary then? :)
just a bit confusing given that it's not part of "normal" usage of the application.

ash2k: Yeah, I thought about making it a separate binary. Let me try to see how it'll look...

pkg/resources/crd_helpers.go
```diff
@@ -68,6 +289,7 @@ func EnsureCrdExists(ctx context.Context, defaulter runtime.ObjectDefaulter, cli
 				return nil
 			}
 			log.Printf("Updating CustomResourceDefinition %s", crd.Name)
+			obj = obj.DeepCopy()
```
ash2k: Was mutating original object from lister 🐛

cmd/smith/app/app.go
```diff
@@ -206,7 +206,7 @@ func NewFromFlags(flagset *flag.FlagSet, arguments []string) (*App, error) {
 	flagset.DurationVar(&a.ResyncPeriod, "resync-period", defaultResyncPeriod, "Resync period for informers")
 	flagset.IntVar(&a.Workers, "workers", 2, "Number of workers that handle events from informers")
 	flagset.StringVar(&a.Namespace, "namespace", meta_v1.NamespaceAll, "Namespace to use. All namespaces are used if empty string or omitted")
-	pprofAddr := flag.String("pprof-address", "", "Address for pprof to listen on")
+	pprofAddr := flagset.String("pprof-address", "", "Address for pprof to listen on")
```
ash2k: 🐛

cmd/smith/app/app.go
ash2k: Looks better now!

pkg/resources/crd_helpers.go
```diff
@@ -32,10 +38,225 @@ func BundleCrd() *apiext_v1b1.CustomResourceDefinition {
 				Singular: smith_v1.BundleResourceSingular,
 				Kind:     smith_v1.BundleResourceKind,
 			},
+			Validation: &apiext_v1b1.CustomResourceValidation{
+				OpenAPIV3Schema: &apiext_v1b1.JSONSchemaProps{
```
wryun: soo... typey?

docs/deployment/0-crd.yaml
```diff
@@ -11,3 +11,147 @@ spec:
     singular: bundle
   scope: Namespaced
   version: v1
+  validation:
```
wryun: why do we have 0-crd.yaml when Smith creates it automatically?

ash2k: It does not anymore https://github.com/atlassian/smith/pull/184

PR #192, opened 2017-12-29
Detect update cycles

### LINKED ISSUE #103: Detect issues in object cleaner
Detect situations where object, returned from update call, does not match the spec and hence notification about update will cause an update again, leading to an infinite loop.
Should catch issues like #101 and prevent hot looping.

ash2k:
Fixes #103. As you can see from the diff, the new code detected issues in existing tests and lack of proper `Secret` handling. 😎

### DIFF
--- a/pkg/cleanup/types/built_in.go
+++ b/pkg/cleanup/types/built_in.go
@@ -16,6 +16,7 @@ import (
 var MainKnownTypes = map[schema.GroupKind]cleanup.SpecCleanup{
 	{Group: apps_v1b2.GroupName, Kind: "Deployment"}: deploymentCleanup,
 	{Group: core_v1.GroupName, Kind: "Service"}:      serviceCleanup,
+	{Group: core_v1.GroupName, Kind: "Secret"}:       secretCleanup,
 }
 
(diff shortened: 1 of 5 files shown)


pkg/cleanup/types/built_in.go
```diff
@@ -63,6 +64,26 @@ func serviceCleanup(spec, actual *unstructured.Unstructured) (*unstructured.Unst
...
+		if secretSpec.Data == nil {
+			secretSpec.Data = make(map[string][]byte, len(secretSpec.StringData))
+		}
+		for k, v := range secretSpec.StringData {
+			secretSpec.Data[k] = []byte(v)
+		}
+		secretSpec.StringData = nil
+	}
+
+	return util.RuntimeToUnstructured(&secretSpec)
```
nilebox: So at this stage the `spec` object will have both `StringData` and `Data` set.
But the `actual` object will only have `Data` set, right? (`StringData` is write-only field). Do you set it somewhere as well, or have some custom comparison logic?

ash2k: Line 81 cleans up `StringData` field.

PR #193, opened 2018-01-01
Leader election

### LINKED ISSUE #119: Do leader election
It is ok to run multiple copies of Smith because it performs all mutations via concurrency safe operations (except deletes - see #118). However there is benefit in avoiding unnecessary load on api servers.
Should use leader election code available in client-go once we are on 5.0. Depends on #111.

ash2k:
Fixes #119.


cmd/smith/app/app.go
```diff
@@ -123,6 +144,29 @@ func (a *App) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	// Events
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(glog.Infof)
+	eventBroadcaster.StartRecordingToSink(&core_v1client.EventSinkImpl{Interface: clientset.CoreV1().Events("")})
+	recorder := eventBroadcaster.NewRecorder(scheme, core_v1.EventSource{Component: "smith-controller"})
```
ash2k: This is here and not in the method because it will be used later to send events.

cmd/smith/app/app.go
```diff
@@ -123,6 +144,29 @@ func (a *App) Run(ctx context.Context) error {
...
 	}
+
+	// Events
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(logInfo)
+	eventBroadcaster.StartRecordingToSink(&core_v1client.EventSinkImpl{Interface: clientset.CoreV1().Events("")})
+	recorder := eventBroadcaster.NewRecorder(scheme, core_v1.EventSource{Component: "smith-controller"})
+
+	// Leader election
+	if a.LeaderElect {
```
wryun: under what situations would you not want this. Does it cause a perf penalty? Also, why not default?

cmd/smith/app/app.go
```diff
@@ -50,6 +70,7 @@ type App struct {
 	Plugins              []plugin.NewFunc
 	Workers              int
 	DisablePodPreset     bool
+	LeaderElectionConfig
```
wryun: why not nest? Makes it easier to understand that a.RetryPeriod etc. is referring to leader election?

cmd/smith/app/app.go
```diff
@@ -207,6 +292,28 @@ func NewFromFlags(flagset *flag.FlagSet, arguments []string) (*App, error) {
...
+		"The interval between attempts by the acting master to renew a leadership slot "+
+		"before it stops leading. This must be less than or equal to the lease duration. "+
+		"This is only applicable if leader election is enabled")
+	flagset.DurationVar(&a.RetryPeriod, "leader-elect-retry-period", defaultRetryPeriod, ""+
+		"The duration the clients should wait between attempting acquisition and renewal "+
+		"of a leadership. This is only applicable if leader election is enabled")
+	flagset.StringVar(&a.ConfigMapNamespace, "leader-elect-configmap-namespace", meta_v1.NamespaceDefault,
+		"Namespace to use for leader election ConfigMap. This is only applicable if leader election is enabled")
+	flagset.StringVar(&a.ConfigMapName, "leader-elect-configmap-name", "smith-leader-elect",
+		"Namespace to use. All namespaces are used if empty string or omitted")
```
wryun: Doco incorrect?

cmd/smith/app/app.go
```diff
@@ -207,6 +292,28 @@ func NewFromFlags(flagset *flag.FlagSet, arguments []string) (*App, error) {
 	flagset.IntVar(&a.Workers, "workers", 2, "Number of workers that handle events from informers")
 	flagset.StringVar(&a.Namespace, "namespace", meta_v1.NamespaceAll, "Namespace to use. All namespaces are used if empty string or omitted")
 	pprofAddr := flagset.String("pprof-address", "", "Address for pprof to listen on")
+	flagset.BoolVar(&a.LeaderElect, "leader-elect", false, ""+
```
wryun: It's worth explaining that this is theoretically not necessary here, I think.

cmd/smith/app/app.go
ash2k: The doc says it is experimental. We can make it an always-on thing once we are confident it works as expected.

cmd/smith/app/app.go
ash2k: Makes sense.

PR #202, opened 2018-01-09
Special case secrets when diff fails

amckague:


pkg/speccheck/speccheck.go
```diff
@@ -110,6 +110,10 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
 	delete(updated.Object, "status")
 
 	if !equality.Semantic.DeepEqual(updated.Object, actualClone.Object) {
+		if (spec.GetKind() == "Secret") {
```
ash2k: This should be also checking the group of the object. Unique object type is identified by group+kind.
Versions evolve but they all represent the same thing withing group+kind e.g. you can read/write the same object using v1 and v1beta1.
Object cleanup works like that - it identifies "cleaner" functions by group+kind. See https://github.com/atlassian/smith/blob/master/pkg/cleanup/object_cleanup.go

pkg/speccheck/speccheck.go
```diff
@@ -110,6 +110,10 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
...
+			log.Printf("Objects are different: Secret object %s has changed", spec.GetName())
```
ash2k: `%q` for names please.

pkg/speccheck/speccheck.go
```diff
@@ -110,7 +111,12 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
 	delete(updated.Object, "status")
 
 	if !equality.Semantic.DeepEqual(updated.Object, actualClone.Object) {
-		log.Printf("Objects are different: %s",
+		gk := spec.GroupVersionKind().GroupKind()
+		if gk.Group == core_v1.GroupName && gk.Kind == "Secret" {
+			log.Printf("Objects are different: Secret object %s has changed", spec.GetName())
+			return updated, false, nil
+		}
+		log.Printf("Objects are different: %q",
```
ash2k: You changed `%s` -> `%q` in the wrong place :D Please change this place back to avoid crazy quoting.

pkg/speccheck/speccheck.go
```diff
@@ -110,6 +111,14 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
 	delete(updated.Object, "status")
 
 	if !equality.Semantic.DeepEqual(updated.Object, actualClone.Object) {
+		gvk := spec.GroupVersionKind()
+		gk := gvk.GroupKind()
+
+		if gvk.Version != "" && gk.Group == core_v1.GroupName && gk.Kind == "Secret" {
```
ash2k: Why are you checking the version? It is always a non-empty string

PR #204, opened 2018-01-09
Port plugin code to manage servceinstances to smith

amckague:
Port collection of smith plugin code into smith proper so all
serviceinstance objects gain the benifit of kicking the updaterequests
field in serviceinstances when the secrets referenced in parametersFrom
changes.


amckague:
I came across this during testing
```
diff --git a/pkg/speccheck/speccheck.go b/pkg/speccheck/speccheck.go
index 106e349..9367656 100644
--- a/pkg/speccheck/speccheck.go
+++ b/pkg/speccheck/speccheck.go
@@ -118,6 +118,10 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
 }
 
 func processAnnotations(spec, actual map[string]string) map[string]string {
+	if actual == nil {
+		actual = make(map[string]string)
+	}
+
 	for key, val := range spec {
 		actual[key] = val
 	}
```
I'm not sure the contract around annotation maps returned in speccheck. I can create a new PR with this to cover cases where we currently panic trying to add to a nil map.

pkg/controller/service_instance_test.go
```diff
@@ -0,0 +1,401 @@
...
+	if err := unstructured_conversion.DefaultConverter.FromUnstructured(spec.Object, &instanceSpec); err != nil {
+		panic(err)
+	}
+	return &instanceSpec
+}
+
+func TestSameChecksumIfNoChanges(t *testing.T) {
+	t.Parallel()
+
+	instanceSpec := sc_v1b1.ServiceInstance{
```
amckague: This can probably be improved

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+)
+
+const (
+	secretParametersChecksumAnnotation = smith.Domain + "/secretParametersChecksum"
+	disabled                           = "disabled"
+)
+
+// works around the fact that service catalog does not know when to send an update request if the
+// parameters section would change as the result of changing a secret referenced in parametersFrom. To
+// do this we record the contents of all referenced secrets in an annotation, compare that annotations value
```
ash2k: Hopefully `the hash of` :hash:

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+// each time a service instance is processed, and force UpdateRequests to a higher value when those secrets
+// change to trigger service catalog to send an update request.
+func (st *resourceSyncTask) forceServiceInstanceUpdates(spec *unstructured.Unstructured, actual runtime.Object, namespace string) (specRet *unstructured.Unstructured, e error) {
+	gvk := spec.GroupVersionKind()
+	gk := spec.GroupVersionKind().GroupKind()
+
+	if gk.Kind == "" || gvk.Version == "" {
```
ash2k: Why check this? This should never happen and is a bug in code.

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+)
+
+// works around the fact that service catalog does not know when to send an update request if the
+// parameters section would change as the result of changing a secret referenced in parametersFrom. To
+// do this we record the contents of all referenced secrets in an annotation, compare that annotations value
+// each time a service instance is processed, and force UpdateRequests to a higher value when those secrets
+// change to trigger service catalog to send an update request.
+func (st *resourceSyncTask) forceServiceInstanceUpdates(spec *unstructured.Unstructured, actual runtime.Object, namespace string) (specRet *unstructured.Unstructured, e error) {
+	gvk := spec.GroupVersionKind()
+	gk := spec.GroupVersionKind().GroupKind()
```
ash2k: This is unnecessary - just read what is needed from `gvk`?

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if gk.Kind == "" || gvk.Version == "" {
+		return spec, nil
+	}
+
+	if gk.Group == sc_v1b1.GroupName && spec.GetKind() == "ServiceInstance" {
```
ash2k: 👍

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return spec, nil
+}
+
+// works by adding using an annotation to remember the previous contents of all the secrets referenced in
+// parametersFrom and forcing updateRequest to higher value when the referenced secrets change. The annotation
+// stores the hex value of the bcryted hash of the sha256 hash of the contents of all secrets referenced in
+// parametersFrom.
+//
+// The annotation is only ever added to the spec object not the actual object. The modified spec is treated as
```
ash2k: This all sounds overcomplicated - why not just always add the annotation, overwriting any existing values?

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+// 2. if there are referenced secrets we proceed to check and generate a new value for the annotation
+//
+// This can be disabled by adding the annotation with the value 'disabled'
+func (st *resourceSyncTask) processServiceInstance(spec *unstructured.Unstructured, actual runtime.Object, namespace string) (specRet *unstructured.Unstructured, e error) {
+	var instanceSpec sc_v1b1.ServiceInstance
+	var previousEncodedChecksum string
+	var updateCount int64
+
+	if err := unstructured_conversion.DefaultConverter.FromUnstructured(spec.Object, &instanceSpec); err != nil {
+		return nil, errors.Wrapf(err, "Failure to parse ServiceInstance spec %q", spec.GetName())
```
ash2k: By convention error messages should start from lowercase letter. Here and below

ash2k: Also, name of the object and/or resource is not needed because the error message is attached to the per-resource condition in the status. Please remove it here and below.

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+		previousEncodedChecksum = spec.ObjectMeta.Annotations[secretParametersChecksumAnnotation]
+		updateCount = spec.Spec.UpdateRequests
+	}
+
+	checkSum, err := st.calculateNewCheckSum(&instanceSpec, namespace, previousEncodedChecksum, updateCount)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to generate new checksum for spec %q", spec.GetName())
+	}
+
+	if previousEncodedChecksum != "" && checkSum != previousEncodedChecksum {
```
ash2k: If `previousEncodedChecksum` is empty we don't know whether we need to increment `updateCount` or not.  So it is safer to not check this and just always increment. Actually, this is confusing. The `if` should be checking  `actual != nil && checkSum != previousEncodedChecksum`. Right?

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+		instanceSpec.Spec.UpdateRequests = updateCount + 1
+	}
+
+	instanceSpec = setAnnotation(instanceSpec, checkSum)
+
+	unstructuredSpec, err := util.RuntimeToUnstructured(&instanceSpec)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to marshal back into unstructed %q", spec.GetName())
```
ash2k: `unstructured`.

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	return unstructuredSpec, nil
+}
+
+func (st *resourceSyncTask) calculateNewCheckSum(instanceSpec *sc_v1b1.ServiceInstance, namespace string, previousEncodedChecksum string, updateCount int64) (string, error) {
```
ash2k: `updateCount` is not used

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+		hash := sha256.Sum256(secretData)
+		_, err = buf.Write(hash[:])
+		if err != nil {
+			return nil, err
+		}
+	}
+	return buf.Bytes(), nil
+}
+
+func setAnnotation(instanceSpec sc_v1b1.ServiceInstance, checksum string) sc_v1b1.ServiceInstance {
```
ash2k: Just take a pointer an mutate in place? No need to copy it. And actually shallow struct copy does not copy the contents of the map anyway so the original map is mutated here (unless it is `nil`).

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (st *resourceSyncTask) calculateNewCheckSum(instanceSpec *sc_v1b1.ServiceInstance, namespace string, previousEncodedChecksum string, updateCount int64) (string, error) {
+	checksumPayload, err := st.generateSecretParametersChecksumPayload(&instanceSpec.Spec, namespace)
+	if err != nil {
+		return "", err
+	}
+
+	if previousEncodedChecksum != "" {
+		previousCheckSum, err := decodeChecksum(previousEncodedChecksum)
```
ash2k: Invalid data in the annotation should not stop the processing - just ignore the error and overwrite the invalid value.

ash2k: Maybe log the error

pkg/controller/service_instance.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return instanceSpec
+}
+
+func generateChecksum(data []byte) ([]byte, error) {
+	return bcrypt.GenerateFromPassword(data, bcrypt.MinCost)
+}
+
+// hashed password with its possible plaintext equivalent, return true if they match
+func validateChecksum(checksum []byte, data []byte) bool {
```
ash2k: Could be written as `checksum, data []byte`

pkg/controller/service_instance_test.go
```diff
@@ -0,0 +1,401 @@
+package controller
+
+import (
```
ash2k:
```
import (
standard library

"github.com/atlassian/smith/......"

everything else
)
```

pkg/controller/service_instance_test.go
```diff
@@ -0,0 +1,401 @@
...
+		},
+	}
+
+	updatedSpec, err := rst.forceServiceInstanceUpdates(spec, nil, defaultNamespace)
+	require.NoError(t, err)
+
+	instanceCheck := serviceInstance(updatedSpec)
+
+	assert.Contains(t, instanceCheck.Annotations, annotaionKey)
+	assert.True(t, instanceCheck.Spec.UpdateRequests == 0, "expected UpdateRequests to be 0 for create")
```
ash2k: `assert.Zero()`

pkg/controller/service_instance_test.go
```diff
@@ -0,0 +1,401 @@
...
+	require.NoError(t, err)
+
+	instanceCheck := serviceInstance(updatedSpec)
+
+	assert.Contains(t, instanceCheck.Annotations, annotaionKey)
+	assert.Equal(t, instanceCheck.Annotations[annotaionKey], "disabled")
+	assert.True(t, instanceCheck.Spec.UpdateRequests == 0, "expected UpdateRequests to be 0 for create")
+}
+
+func TestUpdateInstanceSecrets(t *testing.T) {
```
ash2k: `t.Parallel()`

pkg/controller/service_instance_test.go
```diff
@@ -0,0 +1,401 @@
...
+	instanceSpec := sc_v1b1.ServiceInstance{
+		TypeMeta: meta_v1.TypeMeta{
+			Kind:       "ServiceInstance",
+			APIVersion: sc_v1b1.GroupName + "/v1",
```
ash2k: This does not match the type of the struct (and below too)

ash2k:
Regarding the `processAnnotations()` - yes, this looks like a bug to me. Please send a PR :)

pkg/controller/service_instance.go
amckague: next comment down

pkg/controller/service_instance.go
amckague: yes

ash2k:
@amckague tests are failing to compile ;)

PR #208, opened 2018-01-10
Guard against nil maps for annotations

amckague:


pkg/speccheck/speccheck.go
```diff
@@ -118,6 +118,10 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
 }
 
 func processAnnotations(spec, actual map[string]string) map[string]string {
+	if actual == nil {
+		actual = make(map[string]string)
```
ash2k: Pre-allocate it: `actual = make(map[string]string, len(spec))`

PR #210, opened 2018-01-11
Support loading configuration from a file

ash2k:
This is useful to be able to pick up any cluster configuration in any form kubectl supports.
In practice I did this to see if Docker's built-in Kubernetes support works with our integration tests. It does.
See https://docs.docker.com/docker-for-mac/kubernetes/

I've left a few make targets with references to minikbe. That is because I haven't figured out yet if Service Catalog works on Kube-on-Docker and if aggregation layer is enabled there (it should, it is on by default in 1.8).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ OS = $$(uname -s | tr A-Z a-z)
 BINARY_PREFIX_DIRECTORY=$(OS)_amd64_stripped
 BINARY_PURE_PREFIX_DIRECTORY=$(OS)_amd64_pure_stripped
 #BINARY_RACE_PREFIX_DIRECTORY=$(OS)_amd64_race_stripped
+KUBE_CONTEXT ?= minikube
 
 .PHONY: setup
 setup: setup-ci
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -73,67 +74,59 @@ generate-deepcopy:
...
+integration-test: fmt update-bazel
 	bazel test \
 		--test_env=KUBE_PATCH_CONVERSION_DETECTOR=true \
 		--test_env=KUBE_CACHE_MUTATION_DETECTOR=true \
-		--test_env=KUBERNETES_SERVICE_HOST="$$(minikube ip)" \
-		--test_env=KUBERNETES_SERVICE_PORT=8443 \
-		--test_env=KUBERNETES_CA_PATH="$$HOME/.minikube/ca.crt" \
-		--test_env=KUBERNETES_CLIENT_CERT="$$HOME/.minikube/apiserver.crt" \
-		--test_env=KUBERNETES_CLIENT_KEY="$$HOME/.minikube/apiserver.key" \
+		--test_env=KUBERNETES_CONFIG_FROM=file \
```
amckague: if this is a switch that allows loading from a different place, sometimes, it's probably worth having a test that checks you _can_ still load from envvars.

pkg/client/config.go
```diff
@@ -26,3 +28,30 @@ func ConfigFromEnv() (*rest.Config, error) {
...
+func LoadConfig(configFileFrom, configFileName, configContext string) (*rest.Config, error) {
+	var config *rest.Config
+	var err error
+
+	switch configFileFrom {
+	case "in-cluster":
+		config, err = rest.InClusterConfig()
+	case "environment":
+		config, err = ConfigFromEnv()
+	case "file":
```
amckague: Is there a readme that can be updated to reference how to start smith?

ash2k: No, this is not documented (yet).

PR #211, opened 2018-01-12
Fix Service Catalog integration tests

ash2k:
Also make them work with any cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,15 +84,14 @@ integration-test: fmt update-bazel
 		--test_env=KUBERNETES_CONFIG_CONTEXT=$(KUBE_CONTEXT) \
 		//it:go_default_test
 
-.PHONY: minikube-test-sc
-minikube-test-sc: fmt update-bazel
+.PHONY: integration-test-sc
+integration-test-sc: fmt update-bazel
 	bazel test \
 		--test_env=KUBE_PATCH_CONVERSION_DETECTOR=true \
 		--test_env=KUBE_CACHE_MUTATION_DETECTOR=true \
 		--test_env=KUBERNETES_CONFIG_FROM=file \
 		--test_env=KUBERNETES_CONFIG_FILENAME="$$HOME/.kube/config" \
-		--test_env=KUBERNETES_CONFIG_CONTEXT=minikube \
-		--test_env=SERVICE_CATALOG_URL="http://$$(minikube ip):30080" \
+		--test_env=KUBERNETES_CONFIG_CONTEXT=$(KUBE_CONTEXT) \
 		//it/sc:go_default_test
 
 .PHONY: run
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -84,15 +84,14 @@ integration-test: fmt update-bazel
...
-minikube-test-sc: fmt update-bazel
+.PHONY: integration-test-sc
+integration-test-sc: fmt update-bazel
 	bazel test \
 		--test_env=KUBE_PATCH_CONVERSION_DETECTOR=true \
 		--test_env=KUBE_CACHE_MUTATION_DETECTOR=true \
 		--test_env=KUBERNETES_CONFIG_FROM=file \
 		--test_env=KUBERNETES_CONFIG_FILENAME="$$HOME/.kube/config" \
-		--test_env=KUBERNETES_CONFIG_CONTEXT=minikube \
-		--test_env=SERVICE_CATALOG_URL="http://$$(minikube ip):30080" \
```
ash2k: This is unnecessary with api aggregation.

it/sc/main_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"math/rand"
+	"os"
+	"testing"
+	"time"
+)
+
+func TestMain(m *testing.M) {
+	rand.Seed(time.Now().UnixNano())
```
ash2k: Otherwise random namespace name is not random at all.

it/sc/service_catalog_test.go
```diff
@@ -26,6 +27,13 @@ func TestServiceCatalog(t *testing.T) {
 				ClusterServiceClassExternalName: "user-provided-service",
 				ClusterServicePlanExternalName:  "default",
 			},
+			Parameters: &runtime.RawExtension{
+				Raw: []byte(`{
+	"credentials": {
+		"token": "token"
+	}
```
ash2k: Service Catalog not does validation of the schema and test was failing because parameters is a mandatory thing.

pkg/speccheck/speccheck.go
```diff
@@ -100,7 +101,10 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
 	updated.SetLabels(spec.GetLabels())
 	updated.SetAnnotations(processAnnotations(spec.GetAnnotations(), updated.GetAnnotations()))
 	updated.SetOwnerReferences(spec.GetOwnerReferences()) // TODO Is this ok? Check that there is only one controller and it is THIS bundle
-	updated.SetFinalizers(spec.GetFinalizers())           // TODO Is this ok?
+
+	finalizers := sets.NewString(updated.GetFinalizers()...)
+	finalizers.Insert(spec.GetFinalizers()...)
+	updated.SetFinalizers(finalizers.List()) // Sorted list of all finalizers
```
ash2k: Service Catalog adds a finalizer but we are discarding it here. Spec-vs-actual check after update/create was catching this issue. This shows that nobody ran Service Catalog integration tests for a while - at least since #192.

amckague:
Will we be able to run these tests in travis?

ash2k:
@amckague I tried a while ago with `none` minikube driver but it was not working. Might be worthwhile to try again. Or do something else.

Issue #201, opened 2018-01-09
Explicit secret in Smith should not merge keys


wryun:
If you define an explicit secret in Smith such as:

`
 - name: b
    spec:
      object:
        metadata:
          name: b
        apiVersion: v1
        kind: Secret
        stringData:
          mysecret2: notreallysecret
`
It updates the existing secret (i.e. leaving keys there). This means on resource renames etc. old data is left around.

Given that the usage of explicit secrets is minimal, and this is less of an issue with generated secrets, this is probably low-priority.

PR #214, opened 2018-01-12
Fix secret data merging

ash2k:
Cannot reproduce the issue - any ideas? https://github.com/atlassian/smith/issues/201


wryun:
You're not going to see it in the normal test (it's a behaviour of kube, afaik). And in your integration test, you set 'Data' (not just StringData), so it wipes the old secret.

ash2k:
So the ordering was wrong. Ugh.

PR #220, opened 2018-01-22
bandaid the scannability of error messages

amckague:


pkg/controller/bundle_sync_task.go
```diff
@@ -80,7 +80,7 @@ func (st *bundleSyncTask) process() (retriableError bool, e error) {
 		}
 		_, resErr := resInfo.fetchError()
 		blockedOnError = blockedOnError || resErr != nil
-		log.Printf("[%s/%s] Resource %q, ready: %t, error: %v", st.bundle.Namespace, st.bundle.Name, resName, resInfo.isReady(), resErr)
+		log.Printf("ERROR [%s/%s] Resource %q, ready: %t, error: %v", st.bundle.Namespace, st.bundle.Name, resName, resInfo.isReady(), resErr)
```
ash2k: Put it into `if err != nil` block? Also log when there is no error, without logging `nil` for the error as it does now.

PR #221, opened 2018-01-24
Definitions are not supported in 1.9

ash2k:
Also work around a bug in additional properties support

### DIFF
--- a/docs/deployment/0-crd.yaml
+++ b/docs/deployment/0-crd.yaml
@@ -13,144 +13,131 @@ spec:
   version: v1
   validation:
     openAPIV3Schema: # generated using make print-bundle-schema
-      definitions:
-        DNS_SUBDOMAIN:
-          maxLength: 253
-          minLength: 1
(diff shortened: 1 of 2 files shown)


docs/deployment/0-crd.yaml
```diff
@@ -13,144 +13,131 @@ spec:
   version: v1
   validation:
     openAPIV3Schema: # generated using make print-bundle-schema
-      definitions:
```
ash2k: This stuff below is generated, don't read it.

pkg/resources/crd_helpers.go
```diff
@@ -26,6 +26,157 @@ func BundleCrd() *apiext_v1b1.CustomResourceDefinition {
 	// https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md
 	// https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/namespaces.md
 	// https://github.com/kubernetes/kubernetes/tree/master/api/openapi-spec
+
+	// definitions are not supported, do what we can :)
```
ash2k: Extracted each definition into a variable and use it instead of a reference to that definition.

PR #227, opened 2018-01-31
Handle processing errors more gracefully

ash2k:


pkg/controller/bundle_sync_task.go
```diff
@@ -212,33 +212,48 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
 		readyCond := smith_v1.ResourceCondition{Type: smith_v1.ResourceReady, Status: smith_v1.ConditionFalse}
 		errorCond := smith_v1.ResourceCondition{Type: smith_v1.ResourceError, Status: smith_v1.ConditionFalse}
 
-		resInfo := st.processedResources[res.Name]
```
ash2k: This was causing NPE if the resource was not processed because something failed earlier e.g. the topological sort.

pkg/controller/bundle_sync_task.go
```diff
@@ -212,33 +212,48 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
+				errorCond.Status = smith_v1.ConditionTrue
 				errorCond.Reason = smith_v1.ResourceReasonTerminalError
+				errorCond.Message = fmt.Sprintf("internal error - unknown resource status type %T", resInfo.status)
+				failedResources = append(failedResources, res.Name)
+				retriableResourceErr = false
 			}
-			failedResources = append(failedResources, res.Name)
-			retriableResourceErr = retriableResourceErr && resStatus.isRetriableError // Must not continue if at least one error is not retriable
-		default:
-			panic(errors.Errorf("unknown resource status type %T", resInfo.status))
```
ash2k: Changed this panic to instead be reported as an error on the resource - see above.

pkg/controller/bundle_sync_task.go
```diff
@@ -212,33 +212,48 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
+		} else {
+			// Resource was not processed
+			blockedCond.Status = smith_v1.ConditionUnknown
+			inProgressCond.Status = smith_v1.ConditionUnknown
+			readyCond.Status = smith_v1.ConditionUnknown
+			errorCond.Status = smith_v1.ConditionUnknown
```
nilebox: Do we have enough error reporting in other places in case resource was not processed?
Thinking whether we need to set `errorCond.Status = smith_v1.ConditionTrue` and `errorCond.Message = ...` here.

ash2k: Yes. The only reason a resource might have not been processed if there was an error. That error is passed via `processErr` and is reported via bundle's condition.

PR #230, opened 2018-02-04
Zap Zap Zap!

ash2k:


cmd/crd/main.go
```diff
@@ -14,7 +14,8 @@ import (
 
 func main() {
 	if err := innerMain(); err != nil {
-		log.Fatal(err)
+		fmt.Fprintf(os.Stderr, "%#v", err)
```
nilebox: why don't we use zap everywhere? isn't it initialized yet by this point?

cmd/smith/app/app.go
```diff
@@ -74,13 +74,14 @@ type App struct {
...
 func (a *App) Run(ctx context.Context) error {
+	defer a.Logger.Sync()
 	// Plugins
 	pluginContainers, err := a.loadPlugins()
 	if err != nil {
 		return err
 	}
 	for pluginName := range pluginContainers {
-		log.Printf("Loaded plugin: %q", pluginName)
+		a.Logger.Sugar().Infof("Loaded plugin: %q", pluginName)
```
nilebox: every call on `.Sugar()` creates a new instance of logger, so probably to invoke it only once.
See https://github.com/uber-go/zap/blob/master/logger.go#L126
and https://github.com/uber-go/zap#quick-start

nilebox: Also I don't think the performance of logging is critical for us. Can't we just always use SugaredLogger?

pkg/util/logz/logz.go
```diff
@@ -0,0 +1,45 @@
+package logz
```
nilebox: why `-z` in the package name, tryin' to be cool brah? :)

cmd/crd/main.go
ash2k: No, this is main. What are you going to do if logger initialization fails and you need to print the error?

cmd/smith/app/app.go
ash2k: Yes, performance is not critical, so I call `Sugar()` whenever I feel like 😛 
I like that with structured logging (without sugar) I can use helper functions - see `logz` package. How would I do the same with sugared logger?

pkg/util/logz/logz.go
ash2k: log + zap = logz? If you have a better name for it...?

cmd/smith/app/app.go
ash2k: Also the sugar call is very cheap, so I don't bother keeping the sugared logger around. That is actually how I started, but then sugared logger does not have the nice typed api.

Issue #195, opened 2018-01-02
Reference secrets associated with bindings


wryun:
#194 would be substantially more useful if we could directly reference the secrets output by bindings (which at the moment live outside the bundle, so even if #194 is implemented would not be accessible).

PR #234, opened 2018-02-07
Easy referencing of binding secrets

wryun:
One version of #195, as discussed.

### DIFF
--- a/docs/design/field-references.md
+++ b/docs/design/field-references.md
@@ -30,7 +30,7 @@ spec:
         spec:
           sleepFor: 3
           wakeupMessage: Hello, Infravators!
-        
+
 --- comes later
       status:
         message: "Hello, Infravators!"
@@ -56,4 +56,66 @@ spec:
         metadata:
           name: sleeper3
         spec: "{{{sleeper2#spec}}}"
+```
+
+## Referring to ServiceBinding outputs
+
+When Service Catalog processes a ServiceBinding, the output is placed in a Secret
+(since they might be secret). If they're not secret, it's convenient to directly
+reference them in the bundle. This can be done by using `dependency:bindsecret#Data.secretkey`.
+At the moment, `bindsecret` is the only parameterisation of the dependency that is allowed.
+
+For example:
+
+```yaml
+apiVersion: smith.atlassian.com/v1
+kind: Bundle
+metadata:
+  name: ab
+spec:
+  resources:
+
+  - name: a
+    spec:
+      object:
+        metadata:
+          name: ups-instance
+        apiVersion: servicecatalog.k8s.io/v1beta1
+        kind: ServiceInstance
+        spec:
+          clusterServiceClassExternalName: user-provided-service
+          clusterServicePlanExternalName: default
+          parameters:
+            credentials:
+              password: mypassword
+
+  - name: a-binding
+    dependsOn:
+    - a
+    spec:
+      object:
+        metadata:
+          name: a-binding
+        apiVersion: servicecatalog.k8s.io/v1beta1
+        kind: ServiceBinding
+        spec:
+          instanceRef:
+            name: "{{a#metadata.name}}"
+          secretName: a-binding-secret
+
+  - name: b
+    dependsOn:
+    - a-binding
+    spec:
+      object:
+        metadata:
+          name: ups-instance-2
+        apiVersion: servicecatalog.k8s.io/v1beta1
+        kind: ServiceInstance
+        spec:
+          clusterServiceClassExternalName: user-provided-service
+          clusterServicePlanExternalName: default
+          parameters:
+            x: y
+            password: "{{a-binding:bindsecret#password}}"
 ```
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -96,6 +96,9 @@ const (
 	resSi1 = "resSi1"
 	si1    = "si1"
 	si1uid = "si1-uid"
+	resSi2 = "resSi2"
+	si2    = "si2"
+	si2uid = "si2-uid"
 
 	s1    = "s1"
 	s1uid = "s1-uid"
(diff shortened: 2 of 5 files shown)


pkg/controller/controller_test.go
```diff
@@ -705,6 +708,147 @@ func TestController(t *testing.T) {
 				assert.Equal(t, "specification of the created/updated object does not match the desired spec", resCond.Message)
 			},
 		},
+		"binding secret references are resolved": &testCase{
```
wryun: At least 50% of the time on this PR was trying to figure out this one test... even though I've _already_ looked at them. Might be time to simplify them?

pkg/controller/resource_sync_task.go
```diff
@@ -155,27 +158,56 @@ func (st *resourceSyncTask) processResource(res *smith_v1.Resource) resourceInfo
...
+
+	return resourceInfo{
+		actual:               resUpdated,
+		status:               resourceStatusReady{},
+		serviceBindingSecret: bindingSecret,
+	}
+}
+
+func (st *resourceSyncTask) extractBindingSecret(obj *unstructured.Unstructured) (*core_v1.Secret, error) {
+	actual, err := st.scheme.ConvertToVersion(obj, obj.GroupVersionKind().GroupVersion())
```
wryun: This is calling us to fail CI because the sleeper isn't registered fully with the scheme (or something):

    no kind \"Sleeper\" is registered for version \"crd.atlassian.com/v1\"

I think I should change it to use unstructured OR simply swallow the error (hey, we don't know what type this might be - we might not have registered it with our scheme...), but I'm a little confused here, since I cargo-culted it from the prepareDependencies function in the same file, and it should have the same problem (but not surfaced atm by integration tests). Leaving the breakage here for now for someone who knows more to weigh in.

wryun: Ok, confirmed that it works if you register the scheme, but I don't think we want to do that.

I think I'll change it to check the error code (and the other one).

pkg/controller/controller_test.go
ash2k: Any suggestions?
For readability I think I at least would like to split them into separate files - put each testcase into a separate one into a function that returns a `*testCase`. Then this top level test will just have a map/slice of function invocations to get the test cases.
To avoid cluttering the `controller` directory we can have a sibling `controller_test` package with these tests. It is known as an external test package - it's a special things `go test` knows how to handle. It is a very underdocumented feature. Cannot find a good explanation e.g. it is mentioned here https://dave.cheney.net/2014/12/01/five-suggestions-for-setting-up-a-go-project

> For testing, your test files, those ending with _test.go, may declare themselves to be in the same package, but with _test appended to the package declaration. This is known as an external test. For now, just accept that you can’t put the code for multiple packages into one directory.

pkg/controller/controller_test.go
```diff
@@ -705,6 +708,147 @@ func TestController(t *testing.T) {
...
+			mainClientObjects: []runtime.Object{
+				&core_v1.Secret{
+					ObjectMeta: meta_v1.ObjectMeta{
```
ash2k: I think we also need `TypeMeta`. At least we have it in other tests. It may be that it is no longer needed after my recent refactoring... Separate PR later.

pkg/controller/resource_sync_task.go
ash2k: You are right, I must have introduced a bug in this PR https://github.com/atlassian/smith/pull/228.
No, of course we don't want to register `Sleeper` in our scheme 🤦‍♂️ 
Lets not block this PR on the bug and I'll fix it in a follow up. Sorry for the mess.

pkg/controller/resource_sync_task.go
```diff
@@ -155,27 +158,60 @@ func (st *resourceSyncTask) processResource(res *smith_v1.Resource) resourceInfo
...
 	}
+
+	return resourceInfo{
+		actual:               resUpdated,
+		status:               resourceStatusReady{},
+		serviceBindingSecret: bindingSecret,
+	}
+}
+
+func (st *resourceSyncTask) extractBindingSecret(obj *unstructured.Unstructured) (*core_v1.Secret, error) {
```
ash2k: `maybeExtractBindingSecret()`?

pkg/controller/resource_sync_task.go
ash2k: Also you don't really need to try to convert it to a typed object to check if it is a `ServiceBinding` - just check the GVK. And only then convert. And that will actually help you to not run into the bug :badpokerface:

pkg/controller/spec_processor.go
```diff
@@ -122,18 +131,37 @@ func (sp *SpecProcessor) processMatch(selector string, primitivesOnly bool) (int
...
 	if fieldValue == nil {
 		return nil, errors.Errorf("field not found: %s", selector)
 	}
+
 	if primitivesOnly {
-		switch fieldValue.(type) {
+		switch typedFieldValue := fieldValue.(type) {
+		case []byte:
+			// Secrets are in bytes. We wildly cast them to a string and hope for the best
+			// so we can put them in the JSON in a 'nice' way.
```
ash2k: What will happen if the data is not a valid utf8 sequence? I think json encoder will replace "invalid" bytes with some special character. We can catch it here properly (i.e. check for utf8 validity; there must be a func for this somewhere) and return an error.

docs/design/field-references.md
```diff
@@ -57,3 +57,65 @@ spec:
           name: sleeper3
         spec: "{{{sleeper2#spec}}}"
 ```
+
+## Referring to ServiceBinding outputs
+
+When Service Catalog processes a ServiceBinding, the output is placed in a secret
+(since they might be secret). If they're not secret, it's convenient to directly
+reference them in the bundle. This can be done by using `dependency:bindsecret#secretkey`.
```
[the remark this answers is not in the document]
wryun: Yeah, I was in two minds. I ended up deciding 'bindsecret' implied its own special semantics, so it may as well have enough magic to make the operation as clear as possible (thinking we could add a new magic string if we ever decided we wanted the secret). But, will change.

pkg/controller/spec_processor.go
wryun: Indeed, I was lazy. Will fix, that is definitely better behaviour.

pkg/controller/controller_test.go
wryun: I'm thinking about it, not to much effect.

One thing that I would like to try - although this is not really addressing the complexity issue here - is to structure tests with helpers rather than iterating over maps. So rather than having 'testName: somerandomparams/functions' in a map, it would be:

```
func testName {
  doBlahTest(t, &BlahTestConfig{
    params as before...
  })
  // now do other asserts (i.e. we can more easily drop some of the 'parameters' from above)
}
```

This means you keep the top level test structure (easy testing of a particular test from IDE @nilebox!), and can reasonably add additional asserts without the abstraction of 'some other thing is processing this config and doing something opaque with it'.

pkg/controller/controller_test.go
wryun: There are a bunch of Secrets in this file, and none of them have TypeMeta.

pkg/controller/controller_test.go
ash2k: Cool, that is how our integration tests are structured. That would be much better 👍

wryun:
(the other long term good out of this is that we can remove the secret update checker entirely. Hooray!)

wryun:
Travis build is failing to start kube... (?)

pkg/controller/spec_processor.go
ash2k: The comment does not to be here anymore?

docs/design/field-references.md
```diff
@@ -57,3 +57,65 @@ spec:
           name: sleeper3
         spec: "{{{sleeper2#spec}}}"
 ```
+
+## Referring to ServiceBinding outputs
+
+When Service Catalog processes a ServiceBinding, the output is placed in a Secret
```
ash2k: Could you put backticks around object kinds to make them look consistent with other files?

ash2k:
Travis times out while waiting for minikube to start. Just restart the build in this case.

pkg/controller/spec_processor.go
wryun: I left it there because it's still a bit rude to cast a byte array to a json string just because it works. Might be an accident.

PR #244, opened 2018-02-13
Prereqs for validating service instance

wryun:
Scaffolding (informers) set up, no actual validation yet. Smaller PRs?

Part of #187


cmd/smith/app/app.go
```diff
@@ -117,6 +118,10 @@ func (a *App) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	var catalogger *catalog.Catalog
```
wryun: I'm really unhappy with the naming for this one, but I'm not sure what it should be.

amckague: hmm, var catalogClient *catalog.Client is go ish?

pkg/controller_test/zz_plumbing_for_test.go
```diff
@@ -148,13 +163,49 @@ func (tc *testCase) run(t *testing.T) {
...
+					},
+					Spec: sc_v1b1.ClusterServicePlanSpec{
+						ClusterServiceClassRef: sc_v1b1.ClusterObjectReference{
+							Name: serviceClassName,
+						},
+						ExternalName: servicePlanExternalName,
+						ExternalID:   servicePlanName,
+					},
+				},
+			}...)...,
```
amckague: more dots

wryun: yeah... this was a bit silly. I was a little bit too keen to write it as one expression, I guess.

wryun:
I think I've addressed what we talked about @ash2k , but I probably missed something...

PR #245, opened 2018-02-14
Preprocess references and validate schemas

wryun:
DRAFT - not yet tested. But I'm going away, and feedback would be good by Friday :)


wryun:
Messed up comparison with other branch.

PR #246, opened 2018-02-14
Validate plugins/service instances against schemas

wryun:
When schemas exist and it's possible to fill in default values for references.

Untested, but throwing it out there since I'm away tomorrow.

### DIFF
--- a/docs/design/field-references.md
+++ b/docs/design/field-references.md
@@ -4,11 +4,8 @@ Cross-object field references are a way for an object to get output field(s) of 
 (or multiple dependencies) injected into it as a field (or multiple fields). References to only direct
 dependencies are supported.
 
-Syntax `"{{dependency1#fieldName}}"` means that value of `fieldName` will be injected as a string instead
-of the placeholder. In this case value must be a string, boolean or number.
-
-Syntax `"{{{dependency1#fieldName}}}"` means that value of `fieldName` will be injected without quotes
-instead of the placeholder. In this case it can be of any type including objects.
+Syntax `"{{dependency1#fieldName}}"` means that value of `fieldName` will be injected instead
+of the placeholder. The existing type is maintained.
 
 The `fieldName` could be specified in JsonPath format (with `$.` prefix added by default), for example:
 `{{dependency1#status.conditions[?(@.type=="Ready")].status}}`
@@ -55,7 +52,7 @@ spec:
         kind: Sleeper
         metadata:
           name: sleeper3
-        spec: "{{{sleeper2#spec}}}"
+        spec: "{{sleeper2#spec}}"
 ```
 
 ## Referring to ServiceBinding outputs
@@ -88,6 +85,7 @@ spec:
           parameters:
             credentials:
               password: mypassword
+              host: http://google.com
 
   - name: a-binding
     dependsOn:
@@ -116,6 +114,65 @@ spec:
           clusterServiceClassExternalName: user-provided-service
           clusterServicePlanExternalName: default
           parameters:
-            x: y
+            important: true
+            host: "{{a-binding:bindsecret#host}}"
             password: "{{a-binding:bindsecret#password}}"
 ```
+
+**Warning: it is not currently safe to have a truly secret field
+passed this way (cf `a-binding:bindsecret#password`) unless it's inserted
+into a secret, as it will be exposed in the parameters of the created object.**
+To do the example above correctly, we could instead construct a new secret object
+in the appropriate form for a `ServiceInstance` `parametersFrom` secret reference.
+In future, this [should be automatic](https://github.com/atlassian/smith/issues/233).
+
+## Early validation
+
+Having references inside an object or plugin means that the final
+shape of the object will only be known once the dependencies have
+been evaluated. However, because we would like to fail as quickly
+as possible if the user has entered invalid parameters, 'example'
+values can be specified (as placeholders) so that ServiceInstance
+objects and plugins can be evaluated against their json schemas.
+
+Modifying part of the example from the previous section:
+
+```yaml
+  - name: b
+    dependsOn:
+    - a-binding
+    spec:
+      object:
+        metadata:
+          name: ups-instance-2
+        apiVersion: servicecatalog.k8s.io/v1beta1
+        kind: ServiceInstance
+        spec:
+          clusterServiceClassExternalName: user-provided-service
+          clusterServicePlanExternalName: default
+          parameters:
+            important: true
+            host: '{{a-binding:bindsecret#host#"http://example.com"}}'
+            password: '{{a-binding:bindsecret#password#"fakepassword"}}'
+```
+
+The difference here is that the `host` and `password` fields now have an additional
+`#`, following which there is a chunk of embedded JSON. This allows us
+to validate the parameters against the json schema exposed
+by the OSB catalog endpoint (and currently accessible via a field on
+Service Catalog's ClusterServicePlan resources). Therefore the above
+resource has the provisional `parameters` block for validation purposes
+of:
+
+```json
+{
+  "important": true,
+  "host": "http://example.com",
+  "password": "fakepassword",
+}
+```
+
+This means we can check that `important: true` is reasonable and that we are
+providing all required fields, though of course host/password themselves may
+change. However, if references are used and examples are not provided,
+this validation step is ignored.
--- a/pkg/controller/resource_sync_task.go
+++ b/pkg/controller/resource_sync_task.go
@@ -1,6 +1,8 @@
 package controller
 
 import (
+	"encoding/json"
+
 	"github.com/atlassian/smith"
 	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
 	"github.com/atlassian/smith/pkg/plugin"
@@ -82,6 +84,10 @@ func (st *resourceSyncTask) processResource(res *smith_v1.Resource) resourceInfo
 
 	// Do as much prevalidation of the spec as we can before dependencies are resolved.
 	// (e.g. plugin/service instance/service binding schemas)
+	// We may want to move this out of the resource processing entirely and do
+	// it before we start processing the bundle, both to fail faster and avoid
+	// some unnecessary schema validations (particularly if we want to cache
+	// entire bundle prevalidation results rather than specific validations).
 	if err := st.prevalidate(res); err != nil {
 		return resourceInfo{
 			status: resourceStatusError{
@@ -277,7 +283,10 @@ func (st *resourceSyncTask) getActualObject(res *smith_v1.Resource) (runtime.Obj
 
 // prevalidate does as much validation as possible before doing any real work.
 func (st *resourceSyncTask) prevalidate(res *smith_v1.Resource) error {
+	sp := NewExamplesSpec(res.Name, res.DependsOn)
 	serviceInstanceGvk := sc_v1b1.SchemeGroupVersion.WithKind("ServiceInstance")
+	res = res.DeepCopy() // Spec processor mutates in place
+
 	if res.Spec.Object != nil {
 		if res.Spec.Object.GetObjectKind().GroupVersionKind() == serviceInstanceGvk {
 			if st.catalog == nil {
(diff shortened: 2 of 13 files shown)


pkg/controller/resource_sync_task.go
```diff
@@ -289,23 +296,67 @@ func (st *resourceSyncTask) prevalidate(res *smith_v1.Resource) error {
 				return errors.WithStack(err)
 			}
 			serviceInstance := actual.(*sc_v1b1.ServiceInstance)
```
ash2k: Scheme's conversion methods do not guarantee that a deep copy will be made. And looks like this method mutates the object - should deep copy it first.

pkg/store/catalog.go
```diff
@@ -124,3 +138,101 @@ func (c *Catalog) GetPlanOf(serviceInstanceSpec *sc_v1b1.ServiceInstanceSpec) (*
...
+
+	var schema *gojsonschema.Schema
+	if rawSchema == nil {
+		schema = nil
+	} else {
+		var err error
+		schema, err = gojsonschema.NewSchema(gojsonschema.NewBytesLoader(rawSchema.Raw))
+		if err != nil {
+			return nil, errors.Wrapf(err,
+				"cannot parse json schema for plan %s on broker %s",
```
ash2k: `%q` for names of things please

pkg/store/catalog.go
```diff
@@ -124,3 +138,101 @@ func (c *Catalog) GetPlanOf(serviceInstanceSpec *sc_v1b1.ServiceInstanceSpec) (*
...
+		}
+	}
+
+	c.currentPlanResourceVersions[makePlanResourceVersionKey(plan, action)] = plan.ResourceVersion
+	c.schemas[key] = schema
+	return schema, nil
+}
+
+func (c *Catalog) ValidateServiceInstanceSpec(serviceInstanceSpec *sc_v1b1.ServiceInstanceSpec) error {
+	if serviceInstanceSpec.ParametersFrom != nil && len(serviceInstanceSpec.ParametersFrom) > 0 {
```
ash2k: `len()` is enough

pkg/controller/resource_sync_task.go
```diff
@@ -82,6 +84,9 @@ func (st *resourceSyncTask) processResource(res *smith_v1.Resource) resourceInfo
 
 	// Do as much prevalidation of the spec as we can before dependencies are resolved.
 	// (e.g. plugin/service instance/service binding schemas)
+	// We may want to move this out of the resource processing entirely and do
+	// it before we start processing the bundle, both to fail faster and to avoid
+	// many schema validations.
```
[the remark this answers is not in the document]
wryun: Ah, well, I guess for a naive implementation it mostly helps in the failure case (bundle reprocessing won't happen as many times since the bundle resources won't be changing), but to do it 'properly' I was imagining having a 'ResourceVersion' equivalent (cf the plan caching - would we need an admission controller to set this?) and mark when we've done the validation for that particular bundle. Though ultimately that's not really different from just caching the results of the validation for a particular bundle resource, it's just that we don't have the ugliness of having to compare the parameters block.

But my preference, ultimately, would be to dump the plan cache and instead have only a cache that looking like:

namespace/bundlename/resourcename -> resourceblock/validationresult

Then do a lookup of the resource and a comparison of the resourceblock each time to determine if we can reuse the validation result.

Anyway, as you said, we can solve the problem if it's actually a problem. I'll remove this comment.

pkg/controller/resource_sync_task.go
```diff
@@ -289,23 +296,67 @@ func (st *resourceSyncTask) prevalidate(res *smith_v1.Resource) error {
 				return errors.WithStack(err)
 			}
 			serviceInstance := actual.(*sc_v1b1.ServiceInstance)
-			servicePlan, err := st.catalog.GetPlanOf(&serviceInstance.Spec)
-			if err != nil {
-				return err
+
+			if serviceInstance.Spec.ParametersFrom != nil && len(serviceInstance.Spec.ParametersFrom) > 0 {
+				st.logger.Debug("Not validating against schema due to parametersFrom block", zap.Error(err), zap.String("resourceName", string(res.Name)))
```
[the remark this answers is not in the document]
wryun: Oops, thanks.

pkg/controller/spec_processor.go
```diff
@@ -31,6 +33,16 @@ type SpecProcessor struct {
 	selfName         smith_v1.ResourceName
 	resources        map[smith_v1.ResourceName]*resourceInfo
 	allowedResources map[smith_v1.ResourceName]struct{}
+	defaultsOnly     bool
+}
+
+// errorString is a trivial implementation of error.
```
wryun: err, whoops.

pkg/controller/spec_processor.go
```diff
@@ -125,6 +145,19 @@ func (sp *SpecProcessor) processMatch(selector string, primitivesOnly bool) (int
 	if objName == sp.selfName {
 		return nil, errors.Errorf("self references are not allowed: %s", selector)
 	}
+
+	if sp.defaultsOnly {
+		evalParts := strings.SplitN(parts[1], ResourceOutputSeparator, 2)
```
wryun: this split needs to happen in the normal case too.

pkg/controller/resource_sync_task.go
wryun: Argh, this is really annoying. I wish it was simpler for SpecProcessor to just do a deep copy automatically (guess it would have to use reflection?).

pkg/controller/resource_sync_task.go
```diff
@@ -289,23 +296,67 @@ func (st *resourceSyncTask) prevalidate(res *smith_v1.Resource) error {
...
-			// performance implications of this may be horrifying...
-			_ = servicePlan.Spec.ServiceInstanceCreateParameterSchema
+			if serviceInstance.Spec.Parameters != nil {
+				var parameters map[string]interface{}
+				if err := json.Unmarshal(serviceInstance.Spec.Parameters.Raw, &parameters); err != nil {
+					return errors.Wrapf(err, "unable to unmarshal ServiceInstance resource %q parameters as object", res.Name)
+				}
+
+				if err := sp.ProcessObject(parameters); err != nil {
+					if _, ok := errors.Cause(err).(*noDefaultValueError); ok {
```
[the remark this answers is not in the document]
wryun: I'm not sure about this. I think it's worth people just 'understanding' this one (it's almost a standard Golang cast, and the way the errors package recommends checking this).

wryun:
Still gotta write the tests.

pkg/controller/resource_sync_task.go
ash2k: Spec processor mutates the object - this is by design. But in this particular case we don't want that so need to make a copy. No need for reflectoion, there is a helper method to do "json" copying. `k8s.io/apimachinery/pkg/runtime/DeepCopyJSON()` and `DeepCopyJSONValue()`.

pkg/store/catalog.go
```diff
@@ -124,3 +148,109 @@ func (c *Catalog) GetPlanOf(serviceInstanceSpec *sc_v1b1.ServiceInstanceSpec) (*
...
+	return planSchemaKey(fmt.Sprintf("%s/%s/%s", plan.Name, action))
+}
+
+func (c *Catalog) getSchemaCache(plan *sc_v1b1.ClusterServicePlan, action planSchemaAction) (*gojsonschema.Schema, bool) {
+	key := makePlanSchemaKey(plan, action)
+
+	c.schemasRWMutex.RLock()
+	defer c.schemasRWMutex.RUnlock()
+	if schemaWithRv, ok := c.schemas[key]; ok && schemaWithRv.resourceVersion == plan.ResourceVersion {
+		return schemaWithRv.schema, true
```
ash2k: If the version stored is old we should remove it. Also as a separate PR later we can remove invalidated schemas on update events for plans - the ones that have different resource versions.

pkg/store/catalog.go
```diff
@@ -124,3 +148,109 @@ func (c *Catalog) GetPlanOf(serviceInstanceSpec *sc_v1b1.ServiceInstanceSpec) (*
...
+	}
+	if schema == nil {
+		// no schema to validate against anyway
+		return nil
+	}
+	var parameters []byte
+	if serviceInstanceSpec.Parameters != nil {
+		parameters = serviceInstanceSpec.Parameters.Raw
+	} else {
+		parameters = []byte{}
```
ash2k: Not needed? `nil` slice is a valid slice.

pkg/controller/resource_sync_task.go
wryun: I fixed it to use the deepcopy above before I posted that ;) - it should be fine now. I'm just leery of things that mutate an object in place (I would prefer that SpecProcessor did that by default, but AFAIK DeepCopyJSON would not be sufficient since it could have typed objects, no?)

pkg/store/catalog.go
wryun: I don't think this is necessary - it's safer/easier to reason about if we just 'remove' it by replacing it once we have the new value (see below). Simpler logic, and less risk of unnecessary thrashing around if weird resource processing order.

pkg/store/catalog.go
wryun: schema validator says it doesn't like it. I've made a small change.

wryun:
Main changes: made spec processor less interesting and added tests.

pkg/store/catalog.go
ash2k: Makes sense - we only have a read lock here.

wryun:
Sorry, I'd pushed it to 187-prevalidate-schemas-2 :banana:

pkg/store/catalog.go
```diff
@@ -124,3 +148,109 @@ func (c *Catalog) GetPlanOf(serviceInstanceSpec *sc_v1b1.ServiceInstanceSpec) (*
 		return nil, errors.Errorf("informer reported multiple instances for ClusterServicePlan %q", planKey)
 	}
 }
+
+func makePlanSchemaKey(plan *sc_v1b1.ClusterServicePlan, action planSchemaAction) planSchemaKey {
+	return planSchemaKey(fmt.Sprintf("%s/%s/%s", plan.Name, action))
```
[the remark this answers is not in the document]
wryun: damnit, forgot this. Fixing now.

docs/design/field-references.md
```diff
@@ -88,6 +88,7 @@ spec:
           parameters:
             credentials:
               password: mypassword
+              host: http://google.com
```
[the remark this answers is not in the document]
wryun: Wanted it to be different from the 'example.com' default...

docs/design/field-references.md
```diff
@@ -116,6 +117,65 @@ spec:
...
+        metadata:
+          name: ups-instance-2
+        apiVersion: servicecatalog.k8s.io/v1beta1
+        kind: ServiceInstance
+        spec:
+          clusterServiceClassExternalName: user-provided-service
+          clusterServicePlanExternalName: default
+          parameters:
+            important: true
+            host: "{{a-binding:bindsecret#host#\"http://example.com\"}}"
```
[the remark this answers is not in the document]
wryun: I don't think it's that big an issue, because the defaults shouldn't be human generated.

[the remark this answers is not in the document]
wryun: I've removed the escaping from the example. Hope that helps :P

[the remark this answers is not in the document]
nilebox: If we don't want to make Bundle human readable, I would vote for scrapping the mustache-like format completely, and replacing it with a fully expanded JSON object (not sure myself if kidding or serious). It would be much less ambiguous and confusing than what we have there.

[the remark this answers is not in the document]
wryun: With the 'bundledefaults' object, it's not clear to me how you would cleanly specify this in the autowiring function? Is the idea that each autowiring function would output parallel sets of objects? This seems very heavy-weight. If the main problem is the reference format, we could devolve into something like:

{{smith}}:
  obj: o
  path: x.y
  default: z

Would that be better?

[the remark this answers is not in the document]
nilebox: something like
```
host: >
  {
    "type": "bind-secret-reference",
    "resourceName": "a-binding",
    "path": "host",
    "example": "http://example.com"
  }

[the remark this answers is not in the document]
Comment (wryun) [reply docs/design/field-references.md]: But then how do you know it's a reference without attempted to parse every json object? That's why the json object format suggested above does.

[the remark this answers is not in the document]
Comment (wryun) [reply docs/design/field-references.md]: But yes, example is a nicer word that default in this context, probably.

**Comment by ash2k** [comment]
https://travis-ci.org/atlassian/smith/builds/343643693#L3995
Build failed with
```
--- FAIL: TestServiceInstanceSchemaInvalid (0.12s)
	Error Trace:	service_instance_schema_invalid_test.go:68
			zz_plumbing_for_test.go:343
			service_instance_schema_invalid_test.go:82
	Error:      	Should be false
	Test:       	TestServiceInstanceSchemaInvalid
	Error Trace:	service_instance_schema_invalid_test.go:69
			zz_plumbing_for_test.go:343
			service_instance_schema_invalid_test.go:82
	Error:      	An error is expected but got nil.
	Test:       	TestServiceInstanceSchemaInvalid
	Error Trace:	utils_for_tests.go:21
			service_instance_schema_invalid_test.go:77
			zz_plumbing_for_test.go:343
			service_instance_schema_invalid_test.go:82
	Error:      	Expected value not to be nil.
	Test:       	TestServiceInstanceSchemaInvalid
	Messages:   	resSi1
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x17b5cb3]
goroutine 38 [running]:
testing.tRunner.func1(0xc4204a0ff0)
	GOROOT/src/testing/testing.go:711 +0x5d9
panic(0x1907e00, 0x2381e00)
	GOROOT/src/runtime/panic.go:491 +0x2a2
pkg/controller_test/go_default_xtest_test.TestServiceInstanceSchemaInvalid.func1(0xc4204a0ff0, 0x23af3e0, 0xc42048c360, 0xc4204a1a40, 0xc42023f500, 0xc42048d920)
	pkg/controller_test/service_instance_schema_invalid_test.go:78 +0x463
pkg/controller_test/go_default_xtest_test.(*testCase).run(0xc42023f500, 0xc4204a0ff0)
	pkg/controller_test/zz_plumbing_for_test.go:343 +0x2b9e
pkg/controller_test/go_default_xtest_test.TestServiceInstanceSchemaInvalid(0xc4204a0ff0)
	pkg/controller_test/service_instance_schema_invalid_test.go:82 +0x7f7
testing.tRunner(0xc4204a0ff0, 0x1adb430)
	GOROOT/src/testing/testing.go:746 +0x16d
created by testing.(*T).Run
	GOROOT/src/testing/testing.go:789 +0x569
```

wryun:
Yeah, sorry, I saw the build failure but I was already working on the SNS/SQS stuff, and @nilebox raised some issues that we should discuss. See 'show outdated' on field-references.md.

Issue #102, opened 2017-07-13
Support Generation/ObservedGeneration


ash2k:
Per [recommendation on writing controllers](https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md) Bundle object should have ObservedGeneration in it's status. Some discussions about it:
- https://github.com/kubernetes/kubernetes/issues/7328
- https://github.com/kubernetes/features/issues/95#issuecomment-302946176

It is blocked on splitting spec/status for CustomResources (item in list https://github.com/kubernetes/features/issues/95). Because Smith cannot update status without also bumping up metadata.generation (is it even incremented automatically in 1.7?).

nilebox:
I believe this issue is unblocked now since https://github.com/kubernetes/kubernetes/pull/55168 has just been merged 🎉 
This feature will be supported in the next Kubernetes release - 1.10.

PR #247, opened 2018-02-15
Temporary workaround for foreground deletion in kubectl

nilebox:
- Add a "deleteResources" finalizer to the Bundle
- If the Bundle has a `deletionTimestamp`, perform a "light" foreground deletion logic (delete all resources)
- Once the deletion of resources has succeeded, remove a "deleteResources" finalizer (i.e. proceed with the deletion of the Bundle)


pkg/controller/controller_worker.go
```diff
@@ -108,7 +108,11 @@ func (c *BundleController) ProcessKey(logger *zap.Logger, key string) (retriable
 		scheme:           c.Scheme,
 		catalog:          c.Catalog,
 	}
-	retriable, err := st.process()
+	handled, needsUpdate, retriable, err := st.processForGC()
```
ash2k: As discussed, move the `if` check here.

pkg/controller/gc/finalizer.go
```diff
@@ -0,0 +1,40 @@
+package gc
+
+import (
+	"github.com/atlassian/smith"
+
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	FinalizerDeleteResources string = smith.Domain + "/deleteResources"
```
ash2k: It's better in general to keep constants untyped if they don't have a more specific type. i.e. remove `string`.

pkg/controller/gc/finalizer.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	return false
+}
+
+func HasDeleteResourcesFinalizer(accessor meta_v1.Object) bool {
+	return HasFinalizer(accessor, FinalizerDeleteResources)
+}
+
+func AddDeleteResourcesFinalizer(accessor meta_v1.Object) {
+	accessor.SetFinalizers(append(accessor.GetFinalizers(), FinalizerDeleteResources))
```
ash2k: Should do the same as here probably https://github.com/atlassian/smith/blob/master/pkg/speccheck/speccheck.go#L101-L103
Don't need a set if there is a check for presence already, but sort them for consistency? I'm not sure if order matters at all. Hm...

pkg/controller/bundle_sync_task.go
```diff
@@ -281,100 +282,106 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
-			case resourceStatusReady:
-				readyCond.Status = smith_v1.ConditionTrue
-			case resourceStatusError:
-				errorCond.Status = smith_v1.ConditionTrue
-				errorCond.Message = resStatus.err.Error()
-				if resStatus.isRetriableError {
-					errorCond.Reason = smith_v1.ResourceReasonRetriableError
+	// TODO: We might want to update the status even if bundle is marked for
+	// deletion (e.g. report which resources have been deleted and which block
+	// the deletion of the bundle), but currently it will generate useless "Unknown"
```
amckague: Marking the status as deletion in progress would be useful for reporting tooling. (even if we don't list the resources that could be blocking deletion)

pkg/controller/bundle_sync_task.go
```diff
@@ -385,6 +392,9 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
 	}
 
 	if statusUpdated || finalizersUpdated {
+		bytes, _ := json.Marshal(st.bundle)
```
amckague: left over from testing? This would be a little spammy for info (does zap have debug level?)

pkg/controller/bundle_sync_task.go
```diff
@@ -42,7 +44,13 @@ type bundleSyncTask struct {
 // READY state might mean something different for each resource type. For a Custom Resource it may mean
 // that a field "State" in the Status of the resource is set to "Ready". It is customizable via
 // annotations with some defaults.
-func (st *bundleSyncTask) process() (retriableError bool, e error) {
+func (st *bundleSyncTask) processNormal() (retriableError bool, e error) {
+	// If the "deleteResources" finalizer is missing, add it and finish the processing iteration
+	if !hasDeleteResourcesFinalizer(st.bundle) {
+		st.newFinalizers = addDeleteResourcesFinalizer(st.bundle.GetFinalizers())
```
amckague: Is this racy? Reacting to the delete change at the same time the gc can?

Shouldn't we just add the finalizer to bundles during first create?

pkg/controller/bundle_sync_task.go
```diff
@@ -42,7 +44,13 @@ type bundleSyncTask struct {
...
+		return false, nil
```
amckague: Why do we need to early return here without checking the deleteTimeStamp exists?

nilebox: because we invoke this method only if `deletionTimestamp` is set, see controller_worker.go

pkg/controller/bundle_sync_task.go
nilebox: Yep, sorry leftover from testing.

pkg/controller/bundle_sync_task.go
nilebox: oops, sorry. `processNormal` invoked only if `deletionTimestamp` is **NOT** set. `processDeleted` is invoked otherwise.

amckague: It should be safe to continue processing the bundle without early returning then no?

nilebox: @amckague no, it's not safe:
1. Smith reads a bundle. Adds a finalizer and starts creating resources immediately.
2. User deletes the bundle.
3. Smith finished creating the resources and tries to update the bundle (finalizers and status). It fails, because the version has changed at the step 2.
4. CRD apiserver sees the `deletionTimestamp` and no finalizers and proceeds the deletion.
5. Bundle is gone, but all resources are still there :(

nilebox: In other words, we need to store finalizer in etcd before we do any actions with resources.

amckague: Interesting. I read this as trying hard to prevent users from removing the finalizer rather than just making sure we initialize properly.
Perhaps https://kubernetes.io/docs/admin/extensible-admission-controllers/#what-are-initializers would be closer to this idea.

pkg/controller/bundle_sync_task.go
```diff
@@ -102,6 +109,80 @@ func (st *bundleSyncTask) process() (retriableError bool, e error) {
...
+		ref := objectRef{
+			GroupVersionKind: obj.GetObjectKind().GroupVersionKind(),
+			Name:             m.GetName(),
+		}
+		existingObjs[ref] = m.GetUID()
+	}
+	var firstErr error
+	retriable := true
+	policy := meta_v1.DeletePropagationForeground
+	for ref, uid := range existingObjs {
```
ash2k: What is the reason for two loops?

pkg/controller/resource_sync_task.go
```diff
@@ -504,7 +504,7 @@ func (st *resourceSyncTask) createResource(resClient dynamic.ResourceInterface,
 	if api_errors.IsAlreadyExists(err) {
 		// We let the next ProcessKey() iteration, triggered by someone else creating the resource, to finish the work.
 		err = api_errors.NewConflict(schema.GroupResource{Group: gvk.Group, Resource: gvk.Kind}, spec.GetName(), err)
-		return nil, false, errors.Wrap(err, "object found, but not in Store yet (will re-process)")
+		return nil, false, errors.Wrap(err, "object found, but not in Store yet (will re-processNormal)")
```
ash2k: revert please

pkg/controller/resource_sync_task.go
```diff
@@ -527,7 +527,7 @@ func (st *resourceSyncTask) updateResource(resClient dynamic.ResourceInterface,
 	if err != nil {
 		if api_errors.IsConflict(err) {
 			// We let the next ProcessKey() iteration, triggered by someone else updating the resource, finish the work.
-			return nil, false, errors.Wrap(err, "object update resulted in conflict (will re-process)")
+			return nil, false, errors.Wrap(err, "object update resulted in conflict (will re-processNormal)")
```
ash2k: revert please

pkg/controller/spec_processor.go
```diff
@@ -151,7 +151,7 @@ func (sp *SpecProcessor) processMatch(selector string, primitivesOnly bool) (int
 	jsonPath := fmt.Sprintf("{$.%s}", parts[1])
 	fieldValue, err := resources.GetJsonPathValue(objToTraverse, jsonPath, false)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to process JsonPath reference %s", selector)
+		return nil, errors.Wrapf(err, "failed to processNormal JsonPath reference %s", selector)
```
ash2k: revert please

pkg/controller/spec_processor_test.go
```diff
@@ -154,7 +154,7 @@ func TestSpecProcessorErrors(t *testing.T) {
 			obj: map[string]interface{}{
 				"invalid": "{{{res1#something}}}",
 			},
-			err: `invalid reference at "invalid": failed to process JsonPath reference res1#something: JsonPath execute error: something is not found`,
+			err: `invalid reference at "invalid": failed to processNormal JsonPath reference res1#something: JsonPath execute error: something is not found`,
```
ash2k: revert please

pkg/controller/bundle_sync_task.go
nilebox: @amckague agreed that it's possible to:
- write a dynamic admission controller which adds an initializer
- in the initializer processing add a finalizer and remove the initializer
Not sure if it's worth it, but could be an improvement in the follow-up PR.

pkg/controller/bundle_sync_task.go
nilebox: This method is a subset of `deleteRemovedResources`. Can simplify a bit ofc.

pkg/controller/bundle_sync_task.go
```diff
@@ -199,106 +281,128 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
+
+	if st.newFinalizers != nil {
+		// Update finalizers
+		st.bundle.Finalizers = st.newFinalizers
+		// Set status to "InProgress: True"
+		// (there will be one more iteration for resource sync)
+		// TODO: add more details to the status (reason, resources status)?
+		st.bundle.Status = smith_v1.BundleStatus{
+			Conditions: []smith_v1.BundleCondition{
+				{Type: smith_v1.BundleInProgress, Status: smith_v1.ConditionTrue},
```
amckague: Here is where we'd change the status to something like delete in progress. @ash2k

amckague: @ash2k suggestion is this shouldn't change the status since finalizer changes don't necessarily correspond to operations in progress.

pkg/controller/bundle_sync_task.go
```diff
@@ -199,106 +281,128 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
+				{Type: smith_v1.BundleReady, Status: smith_v1.ConditionFalse},
+				{Type: smith_v1.BundleError, Status: smith_v1.ConditionFalse},
+			},
+		}
+		bundleUpdated = true
+	} else if st.bundle.DeletionTimestamp == nil {
```
amckague: We can cover the else if case for deletiontimestamp != nil and set the status to a new deletioninprogress status.

pkg/controller/bundle_sync_task.go
ash2k: @nilebox status should only be set here if `DeletionTimestamp` is set.

ash2k:
```go
if st.newFinalizers != nil {
  // set finalizers
}
if st.bundle.DeletionTimestamp == nil {
   // bla bla
} else {
   // set in progress
}
```
WDYT?

nilebox: @ash2k @amckague status should change to "InProgress" regardless of `DeletionTimestamp`. Example: user sends an update request which removes finalizers and changes resource list at the same time. Hence the "ready" status which was set before this update is not up-to-date anymore.

nilebox: @amckague so yes, you're right that in reality finalizers change don't necessarily need any updates of the resources, but since we don't have any data about the potential issue I described above, we need to set it to "InProgress" to be safe. The next iteration will correct this status back to "Ready: True" if there's nothing to do.

ash2k: >  Example: user sends an update request which removes finalizers and changes resource list at the same time. Hence the "ready" status which was set before this update is not up-to-date anymore.

@nilebox it will still be not up to date if you set it to anything else than what it should actually be. You are trying to solve the problem that should be solved using generation+observedGeneration (https://github.com/atlassian/smith/issues/102). Also, status may be out of sync with spec [by design](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#spec-and-status):

> When a new version of an object is POSTed or PUT, the "spec" is updated and available immediately. Over time the system will work to bring the "status" into line with the "spec".

Another thing - this code is overwriting the whole status block. It should not. It is potentially erasing useful information about the state of the resource.

nilebox: @ash2k the reason I had to add this status update was a failing `TestUpdate` integration test.
The `AssertBundleTimeout()` method waits until the status is `Ready` and then goes to check if resources are meeting expectations. It turns out that Smith just updated the finalizer and didn't process the resource changes, hence the failing test. Once I introduced an `InProgress` status for finalizer update, the test succeeded. The fix seemed to be appropriate.
Do you think that we need to rewrite the test to resolve this issue instead?

nilebox: Leaving a TODO that we might fix this issue differently in future (using `ObservedGeneration` or something else).

PR #248, opened 2018-02-15
Do nothing if Bundle is marked with DeletionTimestamp

nilebox:
Bugfix to avoid Smith fighting with GC during foreground deletion.

### DIFF
--- a/pkg/controller/controller_worker.go
+++ b/pkg/controller/controller_worker.go
@@ -108,6 +108,12 @@ func (c *BundleController) ProcessKey(logger *zap.Logger, key string) (retriable
 		scheme:           c.Scheme,
 		catalog:          c.Catalog,
 	}
+
+	if st.bundle.DeletionTimestamp != nil {
+		// Nothing to do, let GC do the work
+		return false, nil
+	}
+
 	retriable, err := st.process()
 	return st.handleProcessResult(retriable, err)
 }
--- a/pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
+++ b/pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
@@ -1,0 +1,93 @@
+package controller_test
+
+import (
+	"context"
+	"testing"
+
+	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
+	"github.com/atlassian/smith/pkg/controller"
+	sc_v1b1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	core_v1 "k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	kube_testing "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/cache"
+)
+
+// Should not perform any creates/updates/deletes on resources after bundle
+// is marked with deletionTimestamp and "foregroundDeletion" finalizer
+func TestNoActionsForResourcesWhenForegroundDeletion(t *testing.T) {
+	t.Parallel()
+	now := meta_v1.Now()
+	tc := testCase{
+		mainClientObjects: []runtime.Object{
+			configMapNeedsDelete(),
+			configMapNeedsUpdate(),
+		},
+		scClientObjects: []runtime.Object{
+			serviceInstance(false, false, true),
+		},
+		bundle: &smith_v1.Bundle{
+			ObjectMeta: meta_v1.ObjectMeta{
+				Name:              bundle1,
+				Namespace:         testNamespace,
+				UID:               bundle1uid,
+				DeletionTimestamp: &now,
+				Finalizers:        []string{meta_v1.FinalizerDeleteDependents},
+			},
+			Spec: smith_v1.BundleSpec{
+				Resources: []smith_v1.Resource{
+					{
+						Name: resSi1,
+						Spec: smith_v1.ResourceSpec{
+							Object: &sc_v1b1.ServiceInstance{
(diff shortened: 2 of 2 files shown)


pkg/controller/controller_worker.go
```diff
@@ -108,6 +108,12 @@ func (c *BundleController) ProcessKey(logger *zap.Logger, key string) (retriable
 		scheme:           c.Scheme,
 		catalog:          c.Catalog,
 	}
+
+	if st.bundle.DeletionTimestamp != nil {
```
nilebox: This check can be done before creating a `bundleSyncTask` instance, but since there is a follow-up PR which will handle the `DeletionTimestamp` in a different way, I would prefer to keep it like that.

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
```diff
@@ -0,0 +1,93 @@
...
+		enableServiceCatalog: true,
+		test: func(t *testing.T, ctx context.Context, cntrlr *controller.BundleController, tc *testCase, prepare func(ctx context.Context)) {
+			prepare(ctx)
+			key, err := cache.MetaNamespaceKeyFunc(tc.bundle)
+			require.NoError(t, err)
+			_, err = cntrlr.ProcessKey(tc.logger, key)
+			assert.NoError(t, err)
+
+			actions := tc.bundleFake.Actions()
+			require.Len(t, actions, 2)
```
nilebox: Without this fix there were 3 actions (the last one was updating a bundle).

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
```diff
@@ -0,0 +1,93 @@
...
+									Name: mapNeedsAnUpdate,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		namespace:            testNamespace,
+		enableServiceCatalog: true,
```
amckague: false?

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
```diff
@@ -0,0 +1,93 @@
...
+			prepare(ctx)
+			key, err := cache.MetaNamespaceKeyFunc(tc.bundle)
+			require.NoError(t, err)
+			_, err = cntrlr.ProcessKey(tc.logger, key)
+			assert.NoError(t, err)
+
+			actions := tc.bundleFake.Actions()
+			require.Len(t, actions, 2)
+			_, ok := actions[0].(kube_testing.ListAction)
+			assert.True(t, ok)
```
ash2k: `assert.Implements()` for readability.

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
```diff
@@ -0,0 +1,93 @@
+package controller_test
+
+import (
+	"context"
+	"testing"
+
+	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
+	"github.com/atlassian/smith/pkg/controller"
```
ash2k: Blank line here.

pkg/controller/controller_worker.go
amckague: Should it also be before the building the struct?

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
```diff
@@ -0,0 +1,93 @@
...
+				},
+			},
+		},
+		namespace:            testNamespace,
+		enableServiceCatalog: true,
+		test: func(t *testing.T, ctx context.Context, cntrlr *controller.BundleController, tc *testCase, prepare func(ctx context.Context)) {
+			prepare(ctx)
+			key, err := cache.MetaNamespaceKeyFunc(tc.bundle)
+			require.NoError(t, err)
+			_, err = cntrlr.ProcessKey(tc.logger, key)
```
amckague: we expect the _ to be false here

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
wryun: can't you just add these to the expectedActions list?

pkg/controller/controller_worker.go
nilebox: @amckague which struct are you talking about? if `bundleSyncTask`, have you read my comment? :)

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
nilebox: @wryun thanks for the suggestion, wasn't aware of this feature.

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
nilebox: yes, but I don't think we care about the value of `retriable` when the error is `nil`? Looks like a redundant check to me.

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
nilebox: @wryun hrm, I checked the `actualActions`, and the list is empty. So looks like Bundle actions don't get listed there? @ash2k

pkg/controller/controller_worker.go
amckague: the stuct above. no I didn't check it's name

pkg/controller_test/deleted_bundle_foreground_deletion_noop_test.go
ash2k: @nilebox `actualActions` from `fakeActionHandler` only captures actual HTTP calls made via a dynamic client. All reads for non-CRs happen via typed clients and hence use fake clients in the test. And hence are not recorded where you are looking for them.

PR #252, opened 2018-02-26
Various fixes; ServiceAccount support

ash2k:


pkg/readychecker/types/built_in.go
```diff
@@ -6,19 +6,20 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 var MainKnownTypes = map[schema.GroupKind]readychecker.IsObjectReady{
-	{Group: core_v1.GroupName, Kind: "ConfigMap"}:    alwaysReady,
-	{Group: core_v1.GroupName, Kind: "Secret"}:       alwaysReady,
-	{Group: core_v1.GroupName, Kind: "Service"}:      alwaysReady,
-	{Group: apps_v1b2.GroupName, Kind: "Deployment"}: isDeploymentReady,
-	{Group: ext_v1b1.GroupName, Kind: "Ingress"}:     alwaysReady,
+	{Group: core_v1.GroupName, Kind: "ConfigMap"}:      alwaysReady,
```
wryun: it sort of seems like this should just be something in the config and dynamically generated?
(i.e. a list of object types, regardless of whether they're core, Service Catalog, or whatever...)

wryun: (then ask the cluster if they're supported...)

ash2k: Yes, I know :) One of the future things.

PR #268, opened 2018-04-08
Generic controller

ash2k:
This is a prototype of an idea we came up with @wryun. What if we package all Voyager layers into Smith and execute them in "dry run" mode from a validating admission hook? That way we can **synchronously** validate that all the transformations (by all layers) can be done and that eventual objects match providers' schemas.
This allows for synchronous failures if the user submits an invalid object. If not do it like that then how?

I wanted to see if it is viable to extract common generic machinery and make controllers plugable into it. Looks like it is.

Work left to be done (not this PR?):
- A place to plug an extra set of controllers in
- Validating admission webhook that performs recursive transformation in a generic way (does not know about particular object kinds)

p.s. ignore first two commits. It was a fun weekend :)


pkg/controller/bundlec_test/zz_plumbing_for_test.go
```diff
@@ -217,16 +203,16 @@ func (tc *testCase) run(t *testing.T) {
...
-		catalog, err = store.NewCatalog(serviceClassInf, servicePlanInf)
-		require.NoError(t, err)
 	}
 
-	crdInf := apiext_v1b1inf.NewCustomResourceDefinitionInformer(apiExtClient, 0, cache.Indexers{})
-	bundleInf := client.BundleInformer(smithClient.SmithV1(), meta_v1.NamespaceAll, 0)
+	plugins := make([]plugin.NewFunc, 0, len(tc.plugins))
+	for _, factory := range tc.plugins {
+		pluginInstance := factory(t)
+		plugins = append(plugins, func() (plugin.Plugin, error) { return pluginInstance, nil })
```
nilebox: shouldn't the func implementation invoke `factory(t)` every time? otherwise it's not `NewFunc` (semantically) if it always returns the same instance.

nilebox: ah, it's a test, probably fine.

ash2k:
@nilebox you wouldn't believe, but I've done that already :) see last 15 commits to master. They are the preparation for this refactoring. This cannot be done with less changes - everything needs to change so... I've extracted what I could into separate commits (into master) and split this into 2 main commits.

PR #290, opened 2018-06-04
Use json-based diff instead of reflect-based one in speccheck.CompareActualVsSpec

kopper:
The problem is that diff.ObjectReflectDiff doesn't work well (e.g. panics) when compared objects have completely different schemas (say field x is type of map in A and type of string in B). I can't easily fix this in Kube codebase because this method assumes schema-equality in multiple places. To be fair they check if the types match, they just don't expect the same type can have instances of different shape, like Service Catalog's ServiceInstance can have due to spec.parameters being arbitrary structures (e.g. environment is a string for resource X and map for resource Y).

### DIFF
--- a/pkg/speccheck/speccheck_test.go
+++ b/pkg/speccheck/speccheck_test.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"testing"
 
+	"encoding/json"
+
 	"github.com/atlassian/smith/pkg/cleanup"
 	"github.com/atlassian/smith/pkg/cleanup/types"
 	sc_v1b1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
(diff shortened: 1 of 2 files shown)


ash2k:
Hey, thanks a lot for fixing this! I knew about the issue but... no time to get to it.
Would be nice to have a simple test for `CompareActualVsSpec()` function to ensure we don't regress in the future. You may (or may not) want to add a new test to this file where inequality is exercised https://github.com/atlassian/smith/blob/master/pkg/speccheck/speccheck_test.go

ash2k:
Also something is weird with the commit - it is not attached to your github profile. You may need to attach the email you use to your github profile. Or commit with another email :)

kopper:
Unit test added. I also fixed my Github profile.

pkg/speccheck/speccheck_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"testing"
 
+	"encoding/json"
```
ash2k: Nit: would be good to put it together with other imports. Feel free to merge with or without this change.

PR #291, opened 2018-06-04
gometalinter config and fix some warnings

jokeyrhyme:
- extract duplicated gometalinter configuration out of Makefile into .gometalinter.json
- fix all `make check` warnings: gas, gocyclo, vetshadow
- configure Travis CI to run `make check`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,13 +51,14 @@ install:
 script:
   - make update-bazel
   - make verify
-  - make -j2 check test-ci
+  - make test-ci
 #  - |
 #    until kubectl get nodes -o jsonpath='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}' 2>&1 | grep -q "Ready=True"; do
 #      sleep 1
 #    done
 #  - make integration-test-ci
   - make docker
+  - make check
 
 after_success:
   - if [ "$TRAVIS_OS_NAME" == "linux" -a ! -z "$TRAVIS_TAG" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,10 +49,9 @@ install:
   - make setup-ci
 
 script:
-  - make check
   - make update-bazel
   - make verify
-  - make test-ci
+  - make -j2 check test-ci
 #  - |
 #    until kubectl get nodes -o jsonpath='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}' 2>&1 | grep -q "Ready=True"; do
 #      sleep 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,9 +46,10 @@ before_install:
 #  - minikube update-context
 
 install:
-  - make setup-base
+  - make setup-dev
 
 script:
+  - make check
   - make update-bazel
   - make verify
   - make test-ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ before_install:
 #  - minikube update-context
 
 install:
-  - make setup-dev
+  - make setup-ci
 
 script:
   - make check
--- a/cmd/smith/app/bundle_controller.go
+++ b/cmd/smith/app/bundle_controller.go
@@ -122,19 +122,7 @@ func (c *BundleControllerConstructor) New(config *ctrl.Config, cctx *ctrl.Contex
 
 	var catalog *store.Catalog
 	if c.ServiceCatalogSupport {
-		serviceClassInf, err := svcCatClusterInformer(config, cctx, scClient,
-			sc_v1b1.SchemeGroupVersion.WithKind("ClusterServiceClass"),
-			sc_v1b1inf.NewClusterServiceClassInformer)
-		if err != nil {
-			return nil, err
-		}
-		servicePlanInf, err := svcCatClusterInformer(config, cctx, scClient,
-			sc_v1b1.SchemeGroupVersion.WithKind("ClusterServicePlan"),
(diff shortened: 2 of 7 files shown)


pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -329,50 +329,13 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
-					failedResources = append(failedResources, res.Name)
-					retriableResourceErr = false
-				}
-			} else {
-				// Resource was not processed
-				blockedCond.Status = smith_v1.ConditionUnknown
-				inProgressCond.Status = smith_v1.ConditionUnknown
-				readyCond.Status = smith_v1.ConditionUnknown
-				errorCond.Status = smith_v1.ConditionUnknown
+			blockedCond, inProgressCond, readyCond, errorCond := st.resourceConditions(res)
```
jokeyrhyme: This change helped reduce the cyclomatic complexity for `handleProcessResult()`

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -509,6 +472,58 @@ type objectRef struct {
 	Name string
 }
 
+// resourceConditions calculates conditions for a given Resource,
+// which can be useful when determining whether to retry or not.
+func (st *bundleSyncTask) resourceConditions(res smith_v1.Resource) (
+	blockedCond smith_v1.ResourceCondition,
+	inProgressCond smith_v1.ResourceCondition,
+	readyCond smith_v1.ResourceCondition,
+	errorCond smith_v1.ResourceCondition,
```
ash2k: Please avoid named returns - leave inline comments instead

cmd/smith/app/bundle_controller.go
```diff
@@ -122,17 +122,17 @@ func (c *BundleControllerConstructor) New(config *ctrl.Config, cctx *ctrl.Contex
 
 	var catalog *store.Catalog
 	if c.ServiceCatalogSupport {
-		serviceClassInf, err := svcCatClusterInformer(config, cctx, scClient,
+		serviceClassInf, sciErr := svcCatClusterInformer(config, cctx, scClient,
```
ash2k: I think this makes code much less readable. I don't think this is worth "fixing".

ash2k: And in other places too. Not sure what to do - we cannot enable this linter otherwise, right?

pkg/controller/bundlec/bundle_sync_task.go
ash2k: Nice one! Not because the linter is happy now, but because it reads much better.

cmd/smith/app/bundle_controller.go
jokeyrhyme: It's not a big deal if we don't enable this linter

Even upstream doesn't have this enabled by default, because it generates too many false positives: https://golang.org/src/cmd/vet/shadow.go

From the above linked comments:
> A shadowed variable is a variable declared in an inner scope with the same name and type as a variable in an outer scope, and where the outer variable is mentioned after the inner one
is declared.

So, in this particular case, I think it makes sense to extract this `if` out into a separate function, which would separate the scopes and make shadowing a non-issue

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -164,7 +164,8 @@ func (st *resourceSyncTask) processResource(res *smith_v1.Resource) resourceInfo
 	}
 
 	// Check if resource is ready
-	if ready, retriable, err := st.rc.IsReady(resUpdated); err != nil {
+	var ready bool
+	if ready, retriable, err = st.rc.IsReady(resUpdated); err != nil {
```
jokeyrhyme: One way around vetshadow is to avoid creating new variables with `:=`

.travis.yml
```diff
@@ -46,9 +46,10 @@ before_install:
 #  - minikube update-context
 
 install:
-  - make setup-base
+  - make setup-dev
```
ash2k: Ideally there should be `setup-ci` target for CI.

.travis.yml
```diff
@@ -46,9 +46,10 @@ before_install:
...
 
 script:
+  - make check
```
ash2k: I think this should be the last step, not the first. If my tests are failing I don't want to wait for linter every time. Can we make the linter run concurrently in another job somehow? How long does it take?

jokeyrhyme: It adds ~230 seconds to the build, but I've theoretically made check and test-ci run concurrently

jokeyrhyme:
Hmmm, oddly, Travis CI has a gocyclo error:

> pkg/controller/bundlec/bundle_sync_task.go:296::warning: cyclomatic complexity 33 of function (*bundleSyncTask).handleProcessResult() is high (> 30) (gocyclo)

But when I run `gocyclo -avg pkg/controller/bundlec/bundle_sync_task.go` on my MacBook, it says:

> 28 bundlec (*bundleSyncTask).handleProcessResult pkg/controller/bundlec/bundle_sync_task.go:296:1

ash2k:
I think it would be better to avoid concurrent execution - the logs are mixed together and are hard to read. And also both metalinter and build/tests are CPU-bound and use all cores each. So I doubt we are gaining any speedup.

jokeyrhyme:
I just `git rebase`d my branch on master, this shouldn't have been necessary, but it seems as though Travis CI was merging my branch changes on top of master incorrectly
This should hopefully mean `gocyclo` is running on the same code that I have in my branch, and hopefully should turn the build green :)

jokeyrhyme:
Green!

PR #292, opened 2018-06-04
Plugin statuses

### LINKED ISSUE #287: Report GVKs for used plugins in status
It is impossible to find out group/version/kind of an object produced by a plugin, used in a Bundle by just looking at it. So object's GVK should be reported for each used plugin in the Bundle status.

ash2k:
Fixes #287.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -3,19 +3,20 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 http_archive(
     name = "io_bazel_rules_go",
     sha256 = "c1f52b8789218bb1542ed362c4f7de7052abcf254d865d96fb7ba6d44bc15ee3",
-    url = "https://github.com/bazelbuild/rules_go/releases/download/0.12.0/rules_go-0.12.0.tar.gz",
+    urls = ["https://github.com/bazelbuild/rules_go/releases/download/0.12.0/rules_go-0.12.0.tar.gz"],
 )
 
 http_archive(
     name = "bazel_gazelle",
     sha256 = "ddedc7aaeb61f2654d7d7d4fd7940052ea992ccdb031b8f9797ed143ac7e8d43",
-    url = "https://github.com/bazelbuild/bazel-gazelle/releases/download/0.12.0/bazel-gazelle-0.12.0.tar.gz",
+    urls = ["https://github.com/bazelbuild/bazel-gazelle/releases/download/0.12.0/bazel-gazelle-0.12.0.tar.gz"],
 )
 
 http_archive(
     name = "io_bazel_rules_docker",
+    sha256 = "8333642f283e73818fe14c744ca07ca168b78fe9fe5c7c0e9a37c295a8029b75",
     strip_prefix = "rules_docker-452878d665648ada0aaf816931611fdd9c683a97",
-    url = "https://github.com/bazelbuild/rules_docker/archive/452878d665648ada0aaf816931611fdd9c683a97.zip",
+    urls = ["https://github.com/bazelbuild/rules_docker/archive/452878d665648ada0aaf816931611fdd9c683a97.zip"],
 )
 
 load("@io_bazel_rules_go//go:def.bzl", "go_register_toolchains", "go_rules_dependencies")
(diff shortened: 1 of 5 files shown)


pkg/controller/bundlec_test/zz_plumbing_for_test.go
```diff
@@ -327,6 +328,47 @@ func (tc *testCase) verifyPlugins(t *testing.T) {
...
+	}
+	pluginNames := map[smith_v1.PluginName]struct{}{}
+	for _, res := range tc.bundle.Spec.Resources {
+		if res.Spec.Plugin == nil {
+			continue
+		}
+		pluginNames[res.Spec.Plugin.Name] = struct{}{}
+	}
+	assert.Len(t, bundle.Status.PluginStatuses, len(pluginNames))
+nextPlugin:
```
nilebox: not a fan of goto operator, but for tests fine.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -414,6 +414,41 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
+			}
+			pluginName := res.Spec.Plugin.Name
+			if _, ok := name2status[pluginName]; ok {
+				continue // Already reported
+			}
+			name2status[pluginName] = struct{}{}
+			var pluginStatus smith_v1.PluginStatus
+			pluginContainer, ok := st.pluginContainers[pluginName]
+			if ok {
+				describe := pluginContainer.Plugin.Describe()
```
nilebox: What if plugin returns `nil`?
Probably fine to explode in that case, rather than continue running Smith with invalid plugin.

WORKSPACE
```diff
@@ -3,19 +3,20 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 http_archive(
     name = "io_bazel_rules_go",
     sha256 = "c1f52b8789218bb1542ed362c4f7de7052abcf254d865d96fb7ba6d44bc15ee3",
-    url = "https://github.com/bazelbuild/rules_go/releases/download/0.12.0/rules_go-0.12.0.tar.gz",
+    urls = ["https://github.com/bazelbuild/rules_go/releases/download/0.12.0/rules_go-0.12.0.tar.gz"],
```
nilebox: Where is this change coming from? New version of Bazel? Seems irrelevant to PR.

pkg/apis/smith/v1/types.go
```diff
@@ -149,12 +156,21 @@ func (bc *BundleCondition) String() string {
 	return buf.String()
 }
 
+type PluginStatus struct {
+	Name    PluginName      `json:"name"`
+	Group   string          `json:"group"`
```
nilebox: use `metav1.GroupVersionKind` there (or its copy if you don't want to reuse `metav1` types in API)?

pkg/controller/bundlec/bundle_sync_task.go
ash2k: Hm, I wonder why it is a pointer in the first place? Should probably change it to a non-pointer.
We call this function in a lot of places and never assume it can return a `nil` pointer.

pkg/controller/bundlec_test/zz_plumbing_for_test.go
ash2k: I think this is better than an inline anon function and a `return`.

WORKSPACE
ash2k: This is in another commit, just a cleanup.

pkg/apis/smith/v1/types.go
ash2k: That is what I did initially. But I didn't want to create a sub object for these fields just to re-use the struct.

pkg/controller/bundlec/bundle_sync_task.go
ash2k: I'll send a follow up to remove the pointer here.

ash2k: Or maybe not. Not sure about this one....

PR #294, opened 2018-06-06
gometalinter: golint

jokeyrhyme:
- switch to using gopkg.in/alecthomas/gometalinter.v2 (insulate us slightly from upstream changes)
- enable golint linter, but exclude warnings about comments
- address golint warnings: initialisms in identifiers, unhandled errors, exported identifiers that stutter, underscores in variable names, ALL_CAPS variables, redundant `else`
- where such warnings could not be addressed without breaking compatibility, I left behind a `// TODO: smith.v2: ...` comment, so we can come back to these when we break compatibility for other reasons, if we like

### DIFF
--- a/cmd/smith/main.go
+++ b/cmd/smith/main.go
@@ -13,7 +13,7 @@ import (
 
 func main() {
 	if err := run(); err != nil && err != context.Canceled && err != context.DeadlineExceeded {
-		fmt.Fprintf(os.Stderr, "%#v\n", err)
+		fmt.Fprintf(os.Stderr, "%#v\n", err) // nolint: gas
 		os.Exit(1)
 	}
 }
--- a/examples/sleeper/main/main.go
+++ b/examples/sleeper/main/main.go
@@ -15,7 +15,7 @@ import (
 
 func main() {
 	if err := run(); err != nil && err != context.Canceled && err != context.DeadlineExceeded {
-		fmt.Fprintf(os.Stderr, "%#v", err)
+		fmt.Fprintf(os.Stderr, "%#v", err) // nolint: gas
 		os.Exit(1)
 	}
 }
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -7,32 +7,32 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
-type PluginContainer struct {
+type Container struct {
 	Plugin Plugin
 	schema *gojsonschema.Schema
 }
 
-func NewPluginContainer(newPlugin NewFunc) (PluginContainer, error) {
+func NewContainer(newPlugin NewFunc) (Container, error) {
 	plugin, err := newPlugin()
 	if err != nil {
-		return PluginContainer{}, errors.Wrap(err, "failed to instantiate plugin")
+		return Container{}, errors.Wrap(err, "failed to instantiate plugin")
 	}
 	description := plugin.Describe()
 	var schema *gojsonschema.Schema
 	if description.SpecSchema != nil {
 		schema, err = gojsonschema.NewSchema(gojsonschema.NewBytesLoader(description.SpecSchema))
 		if err != nil {
-			return PluginContainer{}, errors.Wrapf(err, "can't use plugin %q due to invalid schema", description.Name)
+			return Container{}, errors.Wrapf(err, "can't use plugin %q due to invalid schema", description.Name)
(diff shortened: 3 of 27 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+Unreleased
+----------
+- cmd/smith/app: renamed `ApiExtClient` field to `APIExtClient`
+- cmd/smith/app: renamed `ScClient` field to `SCClient`
```
ash2k: We don't provide any code compatibility guarantees so there is no need to call this out.

cmd/smith/main.go
```diff
@@ -13,7 +13,8 @@ import (
 
 func main() {
 	if err := run(); err != nil && err != context.Canceled && err != context.DeadlineExceeded {
-		fmt.Fprintf(os.Stderr, "%#v\n", err)
+		fmt.Fprintf(os.Stderr, "%#v\n", err) // nolint: gas
+		// unhandled error: if we cannot write to stderr, we still exit with code 1 at least
```
ash2k: No need for the comment - code is clear enough :)

examples/sleeper/main/main.go
```diff
@@ -15,7 +15,8 @@ import (
 
 func main() {
 	if err := run(); err != nil && err != context.Canceled && err != context.DeadlineExceeded {
-		fmt.Fprintf(os.Stderr, "%#v", err)
+		fmt.Fprintf(os.Stderr, "%#v", err) // nolint: gas
+		// unhandled error: if we cannot write to stderr, we still exit with code 1 at least
```
ash2k: No need for the comment here too

pkg/plugin/plugin.go
```diff
@@ -7,7 +7,8 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
-type PluginContainer struct {
+// TODO: smith.v2: plugin.PluginContainer stutters; consider calling this Container
```
ash2k: Feel free to just rename, it is ok.

pkg/resources/objects.go
```diff
@@ -9,7 +9,8 @@ import (
 )
 
 // GetJsonPathString extracts the value from the object using given JsonPath template, in a string format
-func GetJsonPathString(obj interface{}, path string) (string, error) {
+// TODO: smith.v2: should be GetJSONPathString
```
ash2k: Here and in other places - just rename =)

CHANGELOG.md
amckague: Do we? The suggested deployment model for smith with plugins is depending directly on the source (and everything that's public in it). In a more general sense its probably good to start putting the effort into updating this when things change.

ash2k: Then we need to define what parts of exported code are considered public API. I also don't want us to slow down because "we cannot make breaking changes". That, of course, does not mean that we cannot start documenting them.

jokeyrhyme:
I made the breaking changes, but I kept everything in CHANGELOG.md as a compromise. If you feel strongly about the contents of CHANGELOG.md, I can make additional changes

CHANGELOG.md
amckague: It's go, everything that starts with an upper case letter is fair game. If that's too much we need some sort of versioned API pkg.

ash2k: That is the thing - I don't think we need to do anything here. We have 20 issues in this project. Each of them is more important than tracking breaking changes at this time (which is work). This is just my opinion, of course.

jokeyrhyme:
Bah, I've tried a few different things, but it seems one linter is having a tough time and is crashing the build:

```
DEBUG: [gotype.52]: executing /home/travis/gopath/bin/gotype -e ./pkg/readychecker/types
DEBUG: [gotype.48]: gotype hits 0: ^(?P<path>.*?\.go):(?P<line>\d+):(?P<col>\d+):\s*(?P<message>.*)$
DEBUG: [gotype.48]: gotype linter took 27.033312788s
DEBUG: [gotype.53]: executing /home/travis/gopath/bin/gotype -e ./pkg/resources
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

I'll come back to this later

PR #297, opened 2018-06-18
Reference names may contain uppercase letters

ash2k:


pkg/resources/crd_helpers.go
```diff
@@ -148,7 +154,7 @@ func BundleCrd() *apiext_v1b1.CustomResourceDefinition {
 		Type:        "object",
 		Required:    []string{"resource"},
 		Properties: map[string]apiext_v1b1.JSONSchemaProps{
-			"name":     DNS_SUBDOMAIN,
+			"name":     referenceName,
```
nilebox: nit: why `DNS_SUBDOMAIN` but `referenceName`? should be either all uppercase, or camel case.

ash2k: This is fixed in another PR https://github.com/atlassian/smith/pull/294/files#diff-9c6e626ae0f128a916b151f9d319b931R31

ash2k: `DNS_SUBDOMAIN` - I was copying stuff from kube docs and forgot about naming conventions :)

PR #299, opened 2018-06-26
Support external ids for classes and plans

ash2k:


pkg/store/catalog.go
```diff
@@ -77,80 +87,109 @@ func serviceClassAndPlanExternalNameIndexKey(serviceClassName string, servicePla
 }
 
 func (c *Catalog) GetClassOf(serviceInstanceSpec *sc_v1b1.ServiceInstanceSpec) (*sc_v1b1.ClusterServiceClass, error) {
-	if serviceInstanceSpec.ClusterServiceClassName == "" && serviceInstanceSpec.ClusterServiceClassExternalName == "" {
-		return nil, errors.Errorf("ServiceInstance must have at least ClusterServiceClassName or ClusterServiceExternalName")
-	}
-	if serviceInstanceSpec.ClusterServiceClassName != "" && serviceInstanceSpec.ClusterServiceClassExternalName != "" {
-		// Not sure if this is true. Maybe ok if they match? But silly.
-		return nil, errors.Errorf("ServiceInstance must have only one of ClusterServiceClassName or ClusterServiceExternalName")
+	switch {
```
nilebox: I find this code a bit confusing to read, but probably okay. Having a comment for what each switch case here and below means would be nice.

Issue #276, opened 2018-04-17
Rework error/status reporting for resources


ash2k:
Currently some things that aren't error are propagated as if they are within the code. For example, if an object, described in a bundle, exists but is marked for deletion that is reported as a terminal error in the resource's status. This is not an error, this should be reflected as "in progress" situation. There are more things like that. In general, more things should be propagated as just information about a resource, not as an error return value. Error return values should be used for true errors like unexpected failures, I/O failures, so on.

PR #300, opened 2018-06-28
Added prometheus metrics around bundle condition transitions

ychen-atlassian:
This adds a bunch of metrics into prometheus endpoint about bundle transitions, which would be useful for alerting purposes.

Also proposal for the same PR: How about expanding the error enums?

E.g. for resources we often just spit out the error but it would be useful to categorize if the Spec was wrong or service catalog didn't respond or if the resource couldn't be created.

```
	spec, err := st.evalSpec(res, actual)
	if err != nil {
		return resourceInfo{
			status: resourceStatusError{
 /// this thing could have something like "ResourceSpecInvalidError"?
				err: err,
			},
		}

```

### DIFF
--- a/pkg/controller/bundlec/bundle_sync_task.go
+++ b/pkg/controller/bundlec/bundle_sync_task.go
@@ -583,6 +583,11 @@ func (st *bundleSyncTask) updateBundleCondition(b *smith_v1.Bundle, condition *s
 
 	if oldCondition == nil {
 		// New resource condition
+		if condition.Status == smith_v1.ConditionTrue {
+			st.bundleTransitionCounter.
+				WithLabelValues(st.bundle.GetNamespace(), st.bundle.GetName(), string(condition.Type), condition.Reason).
+				Inc()
+		}
 		return true
 	}
 
@@ -621,6 +626,11 @@ func (st *bundleSyncTask) updateResourceCondition(b *smith_v1.Bundle, resName sm
 
 	if status == nil {
 		// No status for this resource, hence it's a new resource condition
+		if condition.Status == smith_v1.ConditionTrue {
+			st.bundleResourceTransitionCounter.
+				WithLabelValues(st.bundle.GetNamespace(), st.bundle.GetName(), string(resName), string(condition.Type), condition.Reason).
+				Inc()
+		}
 		return true
 	}
 
@@ -629,6 +639,11 @@ func (st *bundleSyncTask) updateResourceCondition(b *smith_v1.Bundle, resName sm
 
 	if oldCondition == nil {
 		// New resource condition
+		if condition.Status == smith_v1.ConditionTrue {
+			st.bundleResourceTransitionCounter.
+				WithLabelValues(st.bundle.GetNamespace(), st.bundle.GetName(), string(resName), string(condition.Type), condition.Reason).
+				Inc()
+		}
 		return true
 	}
 
--- a/pkg/controller/bundlec/bundle_sync_task.go
+++ b/pkg/controller/bundlec/bundle_sync_task.go
@@ -15,6 +15,7 @@ import (
 	"github.com/atlassian/smith/pkg/util/graph"
 	"github.com/atlassian/smith/pkg/util/logz"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 	api_errors "k8s.io/apimachinery/pkg/api/errors"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -336,10 +340,34 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
 			}
 
-			bundleUpdated = updateResourceCondition(st.bundle, res.Name, &blockedCond) || bundleUpdated
-			bundleUpdated = updateResourceCondition(st.bundle, res.Name, &inProgressCond) || bundleUpdated
-			bundleUpdated = updateResourceCondition(st.bundle, res.Name, &readyCond) || bundleUpdated
-			bundleUpdated = updateResourceCondition(st.bundle, res.Name, &errorCond) || bundleUpdated
+			blockedUpdated := updateResourceCondition(st.bundle, res.Name, &blockedCond) || bundleUpdated
+			inProgressUpdated := updateResourceCondition(st.bundle, res.Name, &inProgressCond) || bundleUpdated
+			readyUpdated := updateResourceCondition(st.bundle, res.Name, &readyCond) || bundleUpdated
+			errorUpdated := updateResourceCondition(st.bundle, res.Name, &errorCond) || bundleUpdated
```
ash2k: Remove ` || bundleUpdated` here and above.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -373,9 +401,27 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
 			}
 		}
 
-		bundleUpdated = updateBundleCondition(st.bundle, &inProgressCond) || bundleUpdated
-		bundleUpdated = updateBundleCondition(st.bundle, &readyCond) || bundleUpdated
-		bundleUpdated = updateBundleCondition(st.bundle, &errorCond) || bundleUpdated
+		inProgressUpdated := updateBundleCondition(st.bundle, &inProgressCond) || bundleUpdated
+		readyUpdated := updateBundleCondition(st.bundle, &readyCond) || bundleUpdated
+		errorUpdated := updateBundleCondition(st.bundle, &errorCond) || bundleUpdated
```
ash2k: Remove `|| bundleUpdated` here and above.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -373,9 +401,27 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
+		if readyUpdated && readyCond.Status == smith_v1.ConditionTrue {
+			st.bundleTransitionCounter.
+				WithLabelValues(st.bundle.GetNamespace(), st.bundle.GetName(), string(readyCond.Type), readyCond.Reason).
+				Inc()
+		}
+		if errorUpdated && errorCond.Status == smith_v1.ConditionTrue {
+			st.bundleTransitionCounter.
+				WithLabelValues(st.bundle.GetNamespace(), st.bundle.GetName(), string(errorCond.Type), errorCond.Reason).
+				Inc()
+		}
```
ash2k: I think all these `if`s can be moved to `updateBundleCondition()`, right? Same with `updateResourceCondition()`.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -373,9 +401,27 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
+			st.bundleTransitionCounter.
+				WithLabelValues(st.bundle.GetNamespace(), st.bundle.GetName(), string(inProgressCond.Type), inProgressCond.Reason).
+				Inc()
+		}
+		if readyUpdated && readyCond.Status == smith_v1.ConditionTrue {
+			st.bundleTransitionCounter.
+				WithLabelValues(st.bundle.GetNamespace(), st.bundle.GetName(), string(readyCond.Type), readyCond.Reason).
+				Inc()
+		}
+		if errorUpdated && errorCond.Status == smith_v1.ConditionTrue {
```
ash2k: So what is the logic here? We increment when a condition becomes true. But what does that show? Why not when it becomes false?

pkg/controller/bundlec/bundle_sync_task.go
ychen-atlassian: Oops - vim replacement errors

pkg/controller/bundlec/bundle_sync_task.go
ychen-atlassian: I left a comment on the other PR

ash2k:
> Also proposal for the same PR: How about expanding the error enums?

Forgot to answer. We need to rework how we propagate errors within Smith. We have https://github.com/atlassian/smith/issues/276 for this work, lets do it in another PR.

PR #305, opened 2018-07-17
Consistent code for dealing with SC setting fields

### LINKED ISSUE #304: cleanup functions don't work properly if externalID is changed
instanceSpec.Spec.ExternalID = instanceActual.Spec.ExternalID

^ bad logic - should copy only if Spec is empty. Otherwise, make sure it can handle error properly from Service Catalog on clash (or just detect here?).

wryun:
Fixes #304 

I'm not clear on whether this is an improvement (argh, compile time type checking lost!), but I wanted to treat the fields consistently.

Please consider carefully before merging.


pkg/cleanup/types/built_in.go
```diff
@@ -114,11 +116,10 @@ func scServiceBindingCleanup(spec, actual *unstructured.Unstructured) (runtime.O
 		return nil, err
 	}
 
-	sbSpec.Spec.ExternalID = sbActual.Spec.ExternalID
-	if sbActual.Spec.UserInfo != nil {
-		sbSpec.Spec.UserInfo = sbActual.Spec.UserInfo.DeepCopy()
```
wryun: It's DeepCopyed before it gets here (in speccheck.go), and the ServiceInstance cleanup wasn't doing this (see below).

PS Maybe cleanup wrong name?

pkg/cleanup/types/built_in.go
```diff
@@ -133,36 +134,52 @@ func scServiceInstanceCleanup(spec, actual *unstructured.Unstructured) (runtime.
...
 
-	if instanceSpec.Spec.ClusterServiceClassName != "" &&
-		instanceSpec.Spec.ClusterServiceClassName != instanceActual.Spec.ClusterServiceClassName {
-		return nil, errors.New("clusterServiceClassName has changed when it should be immutable")
-	}
+	instanceSpec.ObjectMeta.Finalizers = instanceActual.ObjectMeta.Finalizers
 
-	if instanceSpec.Spec.ClusterServicePlanName != "" &&
-		instanceSpec.Spec.ClusterServicePlanName != instanceActual.Spec.ClusterServicePlanName {
-		return nil, errors.New("clusterServicePlanName has changed when it should be immutable")
```
wryun: this check is just wrong (it's possible to change the ClusterServicePlanName)

pkg/cleanup/types/built_in.go
```diff
@@ -133,36 +134,52 @@ func scServiceInstanceCleanup(spec, actual *unstructured.Unstructured) (runtime.
 		return nil, err
 	}
 
-	if instanceSpec.Spec.ClusterServiceClassName != "" &&
-		instanceSpec.Spec.ClusterServiceClassName != instanceActual.Spec.ClusterServiceClassName {
-		return nil, errors.New("clusterServiceClassName has changed when it should be immutable")
```
wryun: this is something service-catalog can enforce, not us.

pkg/cleanup/types/built_in.go
```diff
@@ -133,36 +134,52 @@ func scServiceInstanceCleanup(spec, actual *unstructured.Unstructured) (runtime.
...
-		instanceSpec.Spec.ClusterServicePlanName != instanceActual.Spec.ClusterServicePlanName {
-		return nil, errors.New("clusterServicePlanName has changed when it should be immutable")
-	}
+	setEmptyFieldsFromActual(&instanceSpec.Spec, &instanceActual.Spec, []string{
+		"ClusterServiceClassName",
+		"ClusterServiceClassRef",
+		"ClusterServiceClassExternalName",
+		"ClusterServiceClassExternalID",
 
-	if instanceActual.Spec.ClusterServiceClassExternalName == instanceSpec.Spec.ClusterServiceClassExternalName {
```
wryun: It's not clear to me why this logic is different (and why is uses ExternalName rather than ExternalID).

At any rate, I don't think it's necessary - would prefer to keep strange service catalog specific knowledge out of here.

pkg/cleanup/types/built_in.go
```diff
@@ -114,11 +116,10 @@ func scServiceBindingCleanup(spec, actual *unstructured.Unstructured) (runtime.O
 		return nil, err
 	}
 
-	sbSpec.Spec.ExternalID = sbActual.Spec.ExternalID
-	if sbActual.Spec.UserInfo != nil {
-		sbSpec.Spec.UserInfo = sbActual.Spec.UserInfo.DeepCopy()
-	}
-	sbSpec.Status = sbActual.Status
```
wryun: Status is cleaned up in the default behaviour.

amckague:
Is there some context for this change?

wryun:
If you change the externalID in the bundle smith would accept it and not apply it, because in the cleanup it would copy the externalID from the actual object.

I believe this is wrong: the 'cleanup' should only ever take fields from the actual object if they're not set in the bundle, which is what this does (well, at least for serviceinstance/binding - I don't want to tackle the bigger issue, particularly since @ash2k says it will be solved upstream).

amckague:
Added in bloody minded if statement versions of this in a separate change.

amckague:
find old pr with approvals, add code, merge it like no one's watching, hmm..

amckague:
I've thinned the list of fields to just those service catalog will automatically generate. ClusterServiceClassExternalID (and friends) are required from the user the first time they create an object (and class cannot be changed once picked). The ref's versions are something only the controller sets. The userinfo field is managed by service catalog and cannot be changed sensibly without more thought. 

See https://github.com/kubernetes-incubator/service-catalog/releases/tag/v0.1.28 (and in it's code) for the new fields for namespaced brokers, classes, and plans.

wryun:
@amckague are you sure ClusterServiceClassExternalID isn't auto-set by SC if ClusterServiceClassExternalName is set? (can't easily test atm)

amckague:
Class level objects are only ever read from (never written) in service catalog. There are cases where the plan is written to though.

PR #308, opened 2018-08-03
Run gazelle built in "pure" mode

ash2k:
@amckague does it help on nixos?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -51,7 +51,6 @@ load(
     "@io_bazel_rules_docker//go:image.bzl",
     go_image_repositories = "repositories",
 )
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -61,10 +60,12 @@ go_rules_dependencies()
 
 go_register_toolchains()
 
-go_image_repositories()
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
```
amckague: Why the ordering change?

ash2k: To match the recommended way to load gazelle - I thought it may be important. https://github.com/bazelbuild/bazel-gazelle#running-gazelle-with-bazel

PR #309, opened 2018-08-19
Fix documentation about references

### LINKED ISSUE #267: Update documentation
https://github.com/atlassian/smith/pull/262 makes some documentation examples obsolete. And I guess not just examples, also new syntax needs to be explained.

ash2k:
Fixes #267.

### DIFF
--- a/docs/design/field-references.md
+++ b/docs/design/field-references.md
@@ -4,11 +4,16 @@ Cross-object field references are a way for an object to get output field(s) of 
 (or multiple dependencies) injected into it as a field (or multiple fields). References to only direct
 dependencies are supported.
 
-Syntax `"{{dependency1#fieldName}}"` means that value of `fieldName` will be injected instead
-of the placeholder. The existing type is maintained.
-
-The `fieldName` could be specified in JsonPath format (with `$.` prefix added by default), for example:
-`{{dependency1#status.conditions[?(@.type=="Ready")].status}}`
+References are defined as part of resource definition. Each reference should have a `resource` field that specifies
+the name of the referenced resource. A reference may also have:
+- a `name` that can be used as a substitution marker when wrapped in `!{reference-name-here}`. The existing type
+  is maintained. If `name` is not specified or is not used, the reference effectively becomes just an ordering
+  constraint
+- a `path` that can specify a JSON path expression to extract part(s) of the referenced resource
+- an `example` that can specify an example of the value that is extracted using that reference. It is used for schema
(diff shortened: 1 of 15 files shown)


pkg/controller/bundlec_test/resolve_binding_secret_references_test.go
```diff
@@ -131,32 +104,77 @@ func TestResolveBindingSecretReferences(t *testing.T) {
...
-									Kind:       "ServiceInstance",
-									APIVersion: sc_v1b1.SchemeGroupVersion.String(),
+									Kind:       "ConfigMap",
+									APIVersion: core_v1.SchemeGroupVersion.String(),
 								},
 								ObjectMeta: meta_v1.ObjectMeta{
-									Name: si2,
+									Name: mapNeedsAnUpdate,
 									Annotations: map[string]string{
-										"Secret": "{{" + resSb1 + ":bindsecret#Data.mysecret}}",
```
ash2k: This test should have been failing but it was not. Bundle was updated to add the missing finalizer (see around line 64) and the reference was not processed. In fact, the reference format is old here - test test should have been updated when the references refactoring was performed. I've expanded this test to assert the bundle conditions to make sure it does not regress in the future. Also checked if all other tests do have the finalizer specified (there is one that does not intentionally - it tests the the finalizer is added).

docs/design/field-references.md
```diff
@@ -42,27 +49,30 @@ spec:
...
           name: sleeper3
-        spec: "{{sleeper2#spec}}"
+        spec: "!{sleeper2-spec}"
 ```
 
 ## Referring to ServiceBinding outputs
 
-When Service Catalog processes a ServiceBinding, the output is placed in a Secret
+When Service Catalog processes a `ServiceBinding`, the output is placed in a `Secret`
 (since they might be secret). If they're not secret, it's convenient to directly
```
nilebox: "they may contain sensitive data"

nilebox:
@ash2k you need to run `make check` to see linter errors.

docs/design/field-references.md
ash2k: This text is originally from James, so I'm confident it is grammatically correct 😛

wryun: Don't trust me! I am super-prone to making mistakes. I type slightly more quickly than I think in most cases.

But I think Nail here is not suggesting a grammatical change, but one which avoids the confusion of Secret vs secret, which I probably wrote just because it amused me. His version is better!

PR #313, opened 2018-09-07
Update bazel rules

ash2k:


pkg/apis/smith/v1/types.go
```diff
@@ -144,14 +144,14 @@ type BundleCondition struct {
 
 func (bc *BundleCondition) String() string {
 	var buf bytes.Buffer
-	buf.WriteString(string(bc.Type))
-	buf.WriteByte(' ')
-	buf.WriteString(string(bc.Status))
+	buf.WriteString(string(bc.Type))   // nolint: gosec
```
amckague: What's the warning for this line?

ash2k: unhandled error

PR #314, opened 2018-09-17
Add horizontal pod autoscaler support

mcoot:
PR for adding Smith support for the built-in K8s HorizontalPodAutoscaler objects.

Adds:
* Readiness checking function for HPAs
* Adds informer for HPAs to the Bundle Controller


pkg/readychecker/types/built_in.go
```diff
@@ -62,6 +77,49 @@ func isDeploymentReady(obj runtime.Object) (isReady, retriableError bool, e erro
...
+	foundScalingActive := false
+	for _, cond := range hpa.Status.Conditions {
+		switch cond.Type {
+		case hpaAbleToScaleCondition:
+			switch cond.Status {
+			case core_v1.ConditionTrue:
+				foundAbleToScale = true
+			case core_v1.ConditionFalse:
+				// AbleToScale should not be false if the HPA is working, this is a failure
+				return false, false, errors.Errorf("%s: %s", cond.Reason, cond.Message)
```
nilebox: @ash2k is that correct to return a non-retriable error in that case?
```
	// AbleToScale indicates a lack of transient issues which prevent scaling from occurring,
	// such as being in a backoff window, or being unable to access/update the target scale.
	AbleToScale HorizontalPodAutoscalerConditionType = "AbleToScale"
```
In case of backoff this error is recoverable, and updating the target scale could succeed after retrying...
It doesn't affect the lifecycle of the bundle AFAICT, it's mostly about reporting `InProgress: True` with error `Reason: RetriableError` vs `InProgress: False` and error `Reason: TerminalError`.

pkg/readychecker/types/built_in.go
```diff
@@ -62,6 +77,49 @@ func isDeploymentReady(obj runtime.Object) (isReady, retriableError bool, e erro
...
+		case hpaScalingActiveCondition:
+			switch cond.Status {
+			case core_v1.ConditionTrue:
+				foundScalingActive = true
+			case core_v1.ConditionFalse:
+				// It's expected that it will take a little while to get metrics
+				// If it's been stuck in this condition for >3min we assume it's failed
+				now := meta_v1.Now()
+				if cond.LastTransitionTime.Add(3 * time.Minute).Before(now.Time) {
+					return false, false, errors.Errorf("%s: %s", cond.Reason, cond.Message)
```
nilebox: This also seems like a recoverable error?
e.g. if the autoscale controller was down and now is up again, it will be able to scale.

Gopkg.lock
```diff
@@ -3,18 +3,15 @@
 
 [[projects]]
```
amckague: Why did this file need to change?

pkg/readychecker/types/built_in.go
```diff
@@ -1,33 +1,44 @@
...
 	}
 	ServiceCatalogKnownTypes = map[schema.GroupKind]readychecker.IsObjectReady{
 		{Group: sc_v1b1.GroupName, Kind: "ServiceBinding"}:  isScServiceBindingReady,
 		{Group: sc_v1b1.GroupName, Kind: "ServiceInstance"}: isScServiceInstanceReady,
 	}
-	appsV1Scheme = runtime.NewScheme()
-	scV1B1Scheme = runtime.NewScheme()
+	appsV1Scheme    = runtime.NewScheme()
+	scV1B1Scheme    = runtime.NewScheme()
+	autoscalingV2B1 = runtime.NewScheme()
```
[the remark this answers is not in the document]
amckague: you can add new lines to separate sections to avoid the more horrific alignment shifts

Gopkg.lock
ash2k: @mcoot make sure you have the latest `dep` version

pkg/readychecker/types/built_in.go
ash2k: Yes and no. We need to refactor error propagation in Smith. You are correct, that it influences the reason smith reports. But it also retries to re-process things in the error is marked as retriable, which will not help in this case (but is also not very harmful, only extra logs). So perhaps in this case the error should be marked as retriable to be future compatible with what we will hopefully do to surface such conditions. See https://github.com/atlassian/smith/issues/276

pkg/readychecker/types/built_in.go
```diff
@@ -62,6 +77,49 @@ func isDeploymentReady(obj runtime.Object) (isReady, retriableError bool, e erro
...
+				return false, false, errors.Errorf("%s: %s", cond.Reason, cond.Message)
+			case core_v1.ConditionUnknown:
+				// Assume it's still coming up
+			}
+		case hpaScalingActiveCondition:
+			switch cond.Status {
+			case core_v1.ConditionTrue:
+				foundScalingActive = true
+			case core_v1.ConditionFalse:
+				// It's expected that it will take a little while to get metrics
```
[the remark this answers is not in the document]
ash2k: I think this is fine, we can always change it later. Extract that timeout into a constant and add a comment please.

pkg/readychecker/types/built_in.go
nilebox: Yep, I also had a thought conflict between correct status and Smith indefinitely stuck in retry loop.

pkg/readychecker/types/built_in.go
```diff
@@ -1,33 +1,44 @@
...
+	autoscaling_v2b1 "k8s.io/api/autoscaling/v2beta1"
 	core_v1 "k8s.io/api/core/v1"
 	ext_v1b1 "k8s.io/api/extensions/v1beta1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+const (
+	hpaAbleToScaleCondition   = autoscaling_v2b1.HorizontalPodAutoscalerConditionType("AbleToScale")
```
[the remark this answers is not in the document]
mcoot: Fixed

pkg/readychecker/types/built_in.go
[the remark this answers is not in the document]
mcoot: Fixed

pkg/readychecker/types/built_in.go
mcoot: I have made these retriable.

pkg/readychecker/types/built_in.go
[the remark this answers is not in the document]
mcoot: Done.

pkg/readychecker/types/built_in.go
mcoot: Have made this retriable.

Gopkg.lock
mcoot: Fixed.

ash2k:
👍

PR #320, opened 2018-09-24
Correct capitalisation of HorizontalPodAutoscaler kind

mcoot:
Fix capitalisation of HorizontalPodAutoscaler kind so the informer for HPAs is correctly registered.

### DIFF
--- a/cmd/smith/app/bundle_controller.go
+++ b/cmd/smith/app/bundle_controller.go
@@ -256,7 +256,7 @@ func (c *BundleControllerConstructor) resourceInformers(config *ctrl.Config, cct
 		core_v1.SchemeGroupVersion.WithKind("Secret"):                           core_v1inf.NewSecretInformer,
 		core_v1.SchemeGroupVersion.WithKind("ServiceAccount"):                   core_v1inf.NewServiceAccountInformer,
 		apps_v1.SchemeGroupVersion.WithKind("Deployment"):                       apps_v1inf.NewDeploymentInformer,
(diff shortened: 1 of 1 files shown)


cmd/smith/app/bundle_controller.go
```diff
@@ -256,7 +256,7 @@ func (c *BundleControllerConstructor) resourceInformers(config *ctrl.Config, cct
 		core_v1.SchemeGroupVersion.WithKind("Secret"):                           core_v1inf.NewSecretInformer,
 		core_v1.SchemeGroupVersion.WithKind("ServiceAccount"):                   core_v1inf.NewServiceAccountInformer,
 		apps_v1.SchemeGroupVersion.WithKind("Deployment"):                       apps_v1inf.NewDeploymentInformer,
-		autoscaling_v2b1.SchemeGroupVersion.WithKind("HorizontalPodAutoScaler"): autoscaling_v2b1inf.NewHorizontalPodAutoscalerInformer,
+		autoscaling_v2b1.SchemeGroupVersion.WithKind("HorizontalPodAutoscaler"): autoscaling_v2b1inf.NewHorizontalPodAutoscalerInformer,
```
nilebox: Should this be catched by integration tests?
IIRC they don't run in Travis anymore, but probably on the local machine there is still a way?

mcoot: We don't have integration tests for kube compute yet (that is coming soon). However I had successfully deployed on Playground prior to this change so I'm not entirely sure how that didn't catc hit.

ash2k: You should be able to run Smith's integration tests locally or against a cluster.

PR #321, opened 2018-09-24
Better Bundle CRD validation

ash2k:
This requires Kubernetes 1.11+. Only merge after nobody is using 1.10 (Kube for docker is still 1.10.3).

### DIFF
--- a/pkg/resources/crd_helpers.go
+++ b/pkg/resources/crd_helpers.go
@@ -205,8 +205,7 @@ func BundleCrd() *apiext_v1b1.CustomResourceDefinition {
 			Name: smith_v1.BundleResourceName,
 		},
 		Spec: apiext_v1b1.CustomResourceDefinitionSpec{
-			Group:   smith.GroupName,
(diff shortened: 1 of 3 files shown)


pkg/resources/crd_helpers.go
```diff
@@ -80,23 +80,21 @@ func BundleCrd() *apiext_v1b1.CustomResourceDefinition {
...
-				//		Pattern:   "^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$",
-				//	},
-				//},
+				AdditionalProperties: &apiext_v1b1.JSONSchemaPropsOrBool{
+					Schema: &apiext_v1b1.JSONSchemaProps{
+						Type:      "string",
+						MaxLength: int64ptr(63),
+						Pattern:   "^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$",
+					},
+				},
```
ash2k: This works now thanks to https://github.com/kubernetes/kubernetes/pull/62333.

PR #328, opened 2018-09-26
Add missing RBAC for HPA

ash2k:


docs/deployment/2-cluster-wide-access-setup.yaml
```diff
@@ -59,6 +59,17 @@ rules:
   - update
   - delete
 
+- apiGroups:
+  - autoscaling
+  resources:
+  - horizontalpodautoscalers
+  verbs:
+  - list
```
nilebox: You could just set `*` instead of listing all verbs.

ash2k: But we want to list the needed verbs only to prevent unintended permissions. To catch bugs in code, mostly. It also serves as documentation as to what exactly is needed/used.

PR #329, opened 2018-09-26
Do not log when there is no error

ash2k:
Introduced in https://github.com/atlassian/smith/pull/327 🤦‍♂️

### DIFF
--- a/pkg/controller/bundlec/bundle_sync_task.go
+++ b/pkg/controller/bundlec/bundle_sync_task.go
@@ -404,12 +404,14 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error) 
 	}
 
 	if bundleUpdated {
-		ex := st.updateBundle()
-		if processErr == nil {
-			processErr = ex
-			retriable = true
-		} else {
-			st.logger.Error("Error updating Bundle", zap.Error(ex))
+		err := st.updateBundle()
+		if err != nil {
+			if processErr == nil {
+				processErr = err
+				retriable = true
+			} else {
+				st.logger.Error("Error updating Bundle", zap.Error(err))
+			}
 		}
 	}
 


pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -404,12 +404,14 @@ func (st *bundleSyncTask) handleProcessResult(retriable bool, processErr error)
...
-		ex := st.updateBundle()
-		if processErr == nil {
-			processErr = ex
-			retriable = true
-		} else {
-			st.logger.Error("Error updating Bundle", zap.Error(ex))
+		err := st.updateBundle()
+		if err != nil {
+			if processErr == nil {
+				processErr = err
```
nilebox: you could rewrite this to
```go
if processErr != nil {
  // log new error and propagate the original error
  st.logger.Error("Error updating Bundle", zap.Error(err))
  return retriable, processErr
}
return true, err
```

nilebox: Also, we could probably combine 2 error messages, or wrap one into another.
This could is very confusing to read without comments.

ash2k: It is shorter, true, but I like how currently exceptional (`err != nil`) situation is indented and normal control flow is not. https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

nilebox: I would actually change the code in this function to return the error only in the case when status update fails. The code invoking `handleProcessResult` function already has the original `processErr`, and it's up to the invoking code to decide which one to return as the overall queue item processing result IMO.

So in this code I would just do the normal
```
if err != nil {
  return true, err // possibly wrap it with "Error updating Bundle" description
}

// Otherwise
return false, nil
```

nilebox: Currently when `handleProcessResult` function returns an error, it's not clear whether the processing result has failed, or the original error was returned.

ash2k: Makes sense. Feel free to send a PR or I'll do it later myself.

PR #330, opened 2018-09-26
OpenAPI schema for .status

ash2k:


pkg/resources/crd_helpers.go
```diff
@@ -195,6 +195,129 @@ func BundleCrd() *apiext_v1b1.CustomResourceDefinition {
...
+					Schema: &condition,
+				},
+			},
+		},
+	}
+	objectToDelete := apiext_v1b1.JSONSchemaProps{
+		Type:     "object",
+		Required: []string{"group", "version", "kind", "name"},
+		Properties: map[string]apiext_v1b1.JSONSchemaProps{
+			"group": {
```
ash2k: Group can be an empty string - core resource group.

pkg/resources/crd_helpers.go
```diff
@@ -195,6 +195,129 @@ func BundleCrd() *apiext_v1b1.CustomResourceDefinition {
...
+	}
+	pluginStatus := apiext_v1b1.JSONSchemaProps{
+		Type:     "object",
+		Required: []string{"name", "group", "version", "kind"},
+		Properties: map[string]apiext_v1b1.JSONSchemaProps{
+			"name": {
+				Type:      "string",
+				MinLength: int64ptr(1),
+			},
+			"group": {
```
ash2k: Group can be an empty string - core resource group.

Issue #325, opened 2018-09-26
Label inheritance does not play well with metadata.generation


ash2k:
Currently objects get bundle's labels merged into their labels. Lets say we implement Generation/Observed generation support (https://github.com/atlassian/smith/issues/102). Then consider this scenario:
- bundle has label a=b and resource X. Generation=1
- resource X is created with merged labels
- bundle gets status ready=true and observed generation=1

Now an observer can tell whether `.status` is up to date with `.spec` by looking at Generation/Observed generation (checking if they match).

Then bundle's labels are modified. But labels are not under `.spec` and hence generation is not incremented by the api server for this update. Say resource X has not been updated with new labels because the controller has not processed the updated bundle yet. Now the observer has no way (apart from inspecting the object X) to tell whether the `.status` reflects current spec/labels or not - because generation=1 and observed generation is 1 too.

ash2k:
Everything that influences objects produced by a bundle must be under `.spec`. I'm proposing to remove label inheritance as we don't use it atm. We can re-introduce the feature later as part of `.spec`.

PR #331, opened 2018-09-26
/status subresource and observed generation support

### LINKED ISSUE #102: Support Generation/ObservedGeneration

### LINKED ISSUE #323: Make use of /status subresource
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource

This allows to update the `.spec` part without messing with `.status` and vice versa.

ash2k:
Fixes #323.
Fixes #102.

Need to merge https://github.com/atlassian/smith/pull/330 first.

### DIFF
--- a/it/utils_for_tests.go
+++ b/it/utils_for_tests.go
@@ -31,6 +31,7 @@ import (
 	apiExtClientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apiext_v1b1inf "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions/apiextensions/v1beta1"
 	apiext_v1b1list "k8s.io/apiextensions-apiserver/pkg/client/listers/apiextensions/v1beta1"
+	api_errors "k8s.io/apimachinery/pkg/api/errors"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -70,6 +71,11 @@ func (cfg *Config) CreateObject(ctxTest context.Context, obj, res runtime.Object
 }
 
 func (cfg *Config) AwaitBundleCondition(conditions ...watch.ConditionFunc) *smith_v1.Bundle {
+	// Before checking any conditions, ensure that .status is up to date with .spec
+	generation := IsBundleObservedGenerationCond(cfg.Namespace, cfg.Bundle.Name)
+	for i, cond := range conditions {
+		conditions[i] = AndCond(generation, cond)
+	}
 	lw := cache.NewListWatchFromClient(cfg.SmithClient.SmithV1().RESTClient(), smith_v1.BundleResourcePlural, cfg.Namespace, fields.Everything())
 	event, err := cache.ListWatchUntil(20*time.Second, lw, conditions...)
 	require.NoError(cfg.T, err)
@@ -142,22 +148,14 @@ func IsBundleResourceCond(t *testing.T, namespace, name string, resource smith_v
 	}
 }
 
-func IsBundleNewerCond(namespace, name string, resourceVersions ...string) watch.ConditionFunc {
+func IsBundleObservedGenerationCond(namespace, name string) watch.ConditionFunc {
 	return func(event watch.Event) (bool, error) {
 		metaObj := event.Object.(meta_v1.Object)
 		if metaObj.GetNamespace() != namespace || metaObj.GetName() != name {
 			return false, nil
 		}
 		b := event.Object.(*smith_v1.Bundle)
-		for _, rv := range resourceVersions {
-			if b.ResourceVersion == rv {
-				// TODO Should be using Generation here once it is available
-				// https://github.com/kubernetes/kubernetes/issues/7328
-				// https://github.com/kubernetes/features/issues/95
-				return false, nil
-			}
-		}
-		return true, nil
+		return b.Status.ObservedGeneration >= b.Generation, nil
 	}
 }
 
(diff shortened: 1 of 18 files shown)


it/utils_for_tests.go
```diff
@@ -142,22 +148,14 @@ func IsBundleResourceCond(t *testing.T, namespace, name string, resource smith_v
...
 
-func IsBundleNewerCond(namespace, name string, resourceVersions ...string) watch.ConditionFunc {
+func IsBundleObservedGenerationCond(namespace, name string) watch.ConditionFunc {
 	return func(event watch.Event) (bool, error) {
 		metaObj := event.Object.(meta_v1.Object)
 		if metaObj.GetNamespace() != namespace || metaObj.GetName() != name {
 			return false, nil
 		}
 		b := event.Object.(*smith_v1.Bundle)
-		for _, rv := range resourceVersions {
```
ash2k: Die, hack, die.

it/utils_for_tests.go
```diff
@@ -216,6 +214,18 @@ func SetupApp(t *testing.T, bundle *smith_v1.Bundle, serviceCatalog, createBundl
 	require.NoError(t, err)
 
 	defer func() {
+		cli := smithClient.SmithV1().Bundles(cfg.Namespace)
+		b, err := cli.Get(bundle.Name, meta_v1.GetOptions{})
+		if err == nil {
+			if len(b.Finalizers) > 0 {
+				t.Logf("Removing finalizers from Bundle %q", bundle.Name)
+				b.Finalizers = nil
```
ash2k: It's annoying to remove finalizers manually to make the namespace go away if the test exits abruptly.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -293,42 +292,40 @@ func (st *bundleSyncTask) updateBundle() error {
...
 	if st.newFinalizers != nil {
 		// Update finalizers
 		st.bundle.Finalizers = st.newFinalizers
-		// Set status to "InProgress: True"
-		// (there will be one more iteration for resource sync that will set an appropriate status)
-		// TODO: Setting status to "InProgress" here might be unnecessary
-		// (if the Bundle spec wasn't updated during the last update, the status should be kept with no changes)
-		// The better solution could be storing "ObservedGeneration" in the status to reflect
-		// whether the status is up-to-date with the spec or stale
-		st.bundle.Status.Conditions = []cond_v1.Condition{
```
ash2k: We are sending an update here to the resource's body (not `/status`) which, when subresource `/status` is enabled, ignores `.status` changes. So there is no point making any changes to `.status` - they are ignored. See https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -293,42 +292,40 @@ func (st *bundleSyncTask) updateBundle() error {
...
-			{Type: smith_v1.BundleInProgress, Status: cond_v1.ConditionTrue},
-			{Type: smith_v1.BundleReady, Status: cond_v1.ConditionFalse},
-			{Type: smith_v1.BundleError, Status: cond_v1.ConditionFalse},
-		}
-		_, err := st.updateObjectsToDeleteStatus()
```
ash2k: This is also pointless - `.status` changes are ignored now in this codepath.

pkg/controller/bundlec_test/finalizer_added_if_not_present_test.go
```diff
@@ -87,8 +87,6 @@ func TestFinalizerAdded(t *testing.T) {
 			updateBundle := bundleUpdate.GetObject().(*smith_v1.Bundle)
 			// Make sure that the "deleteResources" finalizer was added
 			assert.True(t, resources.HasFinalizer(updateBundle, bundlec.FinalizerDeleteResources))
-
-			tc.assertObjectsToBeDeleted(t, m1)
```
ash2k: When finalizer is added, we don't touch `.status` anymore (see above). So no need to test it.

it/utils_for_tests.go
```diff
@@ -142,22 +148,14 @@ func IsBundleResourceCond(t *testing.T, namespace, name string, resource smith_v
...
-		for _, rv := range resourceVersions {
-			if b.ResourceVersion == rv {
-				// TODO Should be using Generation here once it is available
-				// https://github.com/kubernetes/kubernetes/issues/7328
-				// https://github.com/kubernetes/features/issues/95
-				return false, nil
-			}
-		}
-		return true, nil
+		return b.Status.ObservedGeneration >= b.Generation, nil
```
nilebox: Just FYI: CRDs don't get `Generation` incremented when annotations/labels updated, see https://github.com/kubernetes/kubernetes/issues/67428
I haven't checked if `ResourceVersion` get incremented in that case, but just something to keep in mind.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -293,42 +292,40 @@ func (st *bundleSyncTask) updateBundle() error {
...
+		err := st.updateBundle()
```
nilebox: I'm pretty sure that `status` subresource allows modifying finalizers.
I have recently found this code in Service Catalog https://github.com/kubernetes-incubator/service-catalog/blob/master/pkg/controller/controller_instance.go#L2420 and double checked that it actually works.
So unless CRD implementation is different, you shouldn't need to invoke this update separately from status update.
Also, there should only be one update request per processing, as we all learned the hard way.

it/utils_for_tests.go
ash2k: > Just FYI: CRDs don't get Generation incremented when annotations/labels updated, see kubernetes/kubernetes#67428

Yes, and that is what I would expect. See https://github.com/atlassian/smith/issues/325 and the corresponding PR. Just found these docs https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/customresources-subresources.md#status-behavior and https://github.com/kubernetes/kubernetes/issues/45539

> I haven't checked if ResourceVersion get incremented in that case, but just something to keep in mind.

Yes, it does. It changes when a mutating request touches `/status` or `/` (normal resource path). I tested =)

pkg/controller/bundlec/bundle_sync_task.go
ash2k: > I'm pretty sure that status subresource allows removing finalizers.

But it should not. See the design doc and the issues I linked above.

> So unless CRD implementation is different, you shouldn't need to invoke this update separately from status update.

When adding finalizers we don't perform bundle processing and hence do not update its status.

> Also, there should only be one update request per processing, as we all learned the hard way.

Yes. That is how it was and this PR does not change this behaviour.

ash2k:
The second commit just splits a method into two to simplify it and make the linter happy, no functional changes.

PR #334, opened 2018-09-27
Test plugin error propagation

### LINKED ISSUE #316: Double check that plugin errors are propagated into status

ash2k:
Fixes #316.


pkg/controller/bundlec_test/plugin_schema_invalid_test.go
```diff
@@ -47,7 +47,6 @@ func TestPluginSchemaInvalid(t *testing.T) {
 		},
 		test: func(t *testing.T, ctx context.Context, cntrlr *bundlec.Controller, tc *testCase) {
 			retriable, err := cntrlr.ProcessBundle(tc.logger, tc.bundle)
-			// Sadly, the actual error is not current propagated
```
ash2k: It is, and there is a test right here in the same file below. Old comment.

PR #337, opened 2018-10-03
Dynamic /status subresource support detection

ash2k:


pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -296,11 +297,29 @@ func (st *bundleSyncTask) updateBundle() error {
 }
 
 func (st *bundleSyncTask) updateBundleStatus() error {
+	// ------ BACKWARDS COMPATIBILITY with existing Bundle CRD without /status. REMOVE ONCE NOT NEEDED ------
+	obj, exists, err := st.store.Get(apiext_v1b1.SchemeGroupVersion.WithKind("CustomResourceDefinition"), meta_v1.NamespaceNone, smith_v1.BundleResourceName)
```
wryun: We definitely have perms for this?

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -296,11 +297,29 @@ func (st *bundleSyncTask) updateBundle() error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to get Bundle CRD")
+	}
+	if !exists {
+		return errors.New("Bundle CRD does not exist")
+	}
+	bundleCRD := obj.(*apiext_v1b1.CustomResourceDefinition)
+	if bundleCRD.Spec.Subresources == nil || bundleCRD.Spec.Subresources.Status == nil { // /status subresource is NOT enabled
+		bundleUpdated, err := st.bundleClient.Bundles(st.bundle.Namespace).Update(st.bundle)
```
wryun: I'm slightly suspicious that there's no test here - could just reinstate an old one maybe?

pkg/controller/bundlec/bundle_sync_task.go
ash2k: This is reading an object from an informer's cache. This is not network IO. We do have permissions to read CRDs because we react dynamically to them being added and removed.

PR #345, opened 2018-10-15
Add missing namespace code

amckague:


ash2k:
This does not even compile. How did it pass CI?

PR #346, opened 2018-10-17
Update dependencies to kube 1.12

ash2k:


pkg/speccheck/speccheck.go
```diff
@@ -102,9 +92,73 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
...
+	}
+}
+
+// TODO Is this ok? Check that there is only one controller and it is THIS bundle
+func processOwnerReferences(spec, actual []meta_v1.OwnerReference) []meta_v1.OwnerReference {
+	if len(spec) == 0 {
+		// return nil slice to make it go away
+		return nil
+	}
+	return spec
```
nilebox: Why do we pass `actual` if we never use it?
Is this function properly implemented?

pkg/speccheck/speccheck.go
```diff
@@ -102,9 +92,73 @@ func (sc *SpecCheck) compareActualVsSpec(spec, actual *unstructured.Unstructured
...
+}
+
+// TODO Nukes added labels. Should be configurable per-object and/or per-object kind?
+func processLabels(spec, actual map[string]string) map[string]string {
```
nilebox: `actual` not used

pkg/speccheck/speccheck.go
nilebox: So we need to keep non-controller ownerReferences?

ash2k: I've removed unused arguments.

PR #351, opened 2018-10-23
Move diff logging into the resource sync task

ash2k:


pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -574,6 +575,12 @@ func (st *resourceSyncTask) updateResource(resClient dynamic.ResourceInterface,
 		return updated, false, nil
 	}
 
+	if difference == "" {
+		st.logger.Info("Objects are different: Secret object has changed")
+	} else {
+		st.logger.Sugar().Infof("Objects are different: %s", difference)
+	}
```
ash2k: That way we use the logger with all the fields filled in.

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -574,6 +575,12 @@ func (st *resourceSyncTask) updateResource(resClient dynamic.ResourceInterface,
 		return updated, false, nil
 	}
 
+	if difference == "" {
+		st.logger.Info("Objects are different: Secret object has changed")
```
wryun: Given that difference is just a string, I would be tempted to return just "secret object difference not reported" from the CompareActualVsSpec function, and not have this slightly messy interpretation of an empty string.

ash2k: good point. How did it not occur to me...

PR #356, opened 2018-10-25
Extend ReadyChecker to allow passing more types and more information

ychen-atlassian:
This is the first part of a few changes to clean up the number of errors being logged in smith + distinguishing between user/provider errors from "Smith errors".

This should be a no-op except with some additional changes to clean up error wrapping and logging.


pkg/statuschecker/ready_checker.go
```diff
@@ -0,0 +1,166 @@
...
+	// 3. Check if it is a CRD with Kind/GroupVersion annotation
+	crd, err = c.crdWithKindGroupVersionAnnotation(gk)
+	if err != nil {
+		return nil, err
+	}
+	if crd != nil {
+		return c.checkForInstance(crd, obj)
+	}
+
+	return ObjectStatusInProgress{}, nil
```
ash2k: I think for unknown objects we should either return an error or ready. in progress does not make any sense. I know this is how it works right now, but we should fix that. Would be great as a follow up if you have time.

pkg/statuschecker/types/built_in.go
```diff
@@ -68,31 +69,33 @@ func init() {
...
-func alwaysReady(_ runtime.Object) (isReady, retriableError bool, e error) {
-	return true, false, nil
+func alwaysReady(_ runtime.Object) (r statuschecker.ObjectStatusResult, e error) {
+	return statuschecker.ObjectStatusReady{}, nil
 }
 
 // Works according to https://kubernetes.io/docs/user-guide/deployments/#the-status-of-a-deployment
 // and k8s.io/kubernetes/pkg/client/unversioned/conditions.go:120 DeploymentHasDesiredReplicas()
-func isDeploymentReady(obj runtime.Object) (isReady, retriableError bool, e error) {
+func isDeploymentReady(obj runtime.Object) (r statuschecker.ObjectStatusResult, e error) {
```
ash2k: Please remove named return vales if they are not needed

pkg/statuschecker/types/built_in.go
```diff
@@ -108,7 +111,10 @@ func isHorizontalPodAutoscalerReady(obj runtime.Object) (isReady, retriableError
 				foundAbleToScale = true
 			case core_v1.ConditionFalse:
 				// AbleToScale should not be false if the HPA is working, this is a (retriable) failure
-				return false, true, errors.Errorf("%s: %s", cond.Reason, cond.Message)
+				return statuschecker.ObjectStatusError{
```
ash2k: Nice! Not it is just business as usual, not an error.

ash2k: Worth returning reason and message as separate fields? Both here and in other places... Not sure about this. Can be done later.

pkg/statuschecker/types/built_in.go
```diff
@@ -121,55 +127,94 @@ func isHorizontalPodAutoscalerReady(obj runtime.Object) (isReady, retriableError
...
-	if readyCond != nil && readyCond.Status == sc_v1b1.ConditionTrue {
-		return true, false, nil
+	if readyCond != nil {
+		switch readyCond.Status {
+		case sc_v1b1.ConditionFalse:
+			return statuschecker.ObjectStatusInProgress{
+				Message: fmt.Sprintf("%v: %v", readyCond.Reason, readyCond.Message),
+			}, nil
+		case sc_v1b1.ConditionTrue:
+			return statuschecker.ObjectStatusReady{}, nil
```
ash2k: Might want to add reason+message if they are (or one of them) not empty.

pkg/statuschecker/ready_checker.go
```diff
@@ -0,0 +1,168 @@
...
+	ObjectStatusTypeInProgress ObjectStatusType = "InProgress"
+	ObjectStatusTypeError      ObjectStatusType = "Error"
+)
+
+type ObjectStatusResult interface {
+	StatusType() ObjectStatusType
+}
+
+type ObjectStatusReady struct {
+	Message string
```
ash2k: This is not used anywhere and also is not filled in everywhere (this is fine, but worth double checking if there are places where we could put something in there).

PR #357, opened 2018-10-26
Added deployment checksum annotation to pod template for envFrom updates

ychen-atlassian:
I am not sure if bumping the annotation will cause the pod to be reprocessed. I am going to test smith in playground on monday.


pkg/controller/bundlec/deployment_test.go
```diff
@@ -0,0 +1,422 @@
...
+		Spec: apps_v1.DeploymentSpec{
+			Template: core_v1.PodTemplateSpec{
+				Spec: core_v1.PodSpec{
+					Containers: []core_v1.Container{
+						core_v1.Container{}, // empty EnvFrom
+						core_v1.Container{
+							EnvFrom: []core_v1.EnvFromSource{
+								core_v1.EnvFromSource{
+									SecretRef: &core_v1.SecretEnvSource{
+										LocalObjectReference: core_v1.LocalObjectReference{
```
ychen-atlassian: This thing is kind-of really long.

nilebox:
> I am not sure if bumping the annotation will cause the pod to be reprocessed.

Yes, it should. I saw comments from SIG Architecture recommending this approach in related GitHub issues.
Also see https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#user-content-automatically-roll-deployments-when-configmaps-or-secrets-change

pkg/controller/bundlec/deployment.go
```diff
@@ -0,0 +1,154 @@
...
+				}
+				if !exists {
+					return nil, errors.Errorf("%v %q referenced from envFrom not found", gvk.Kind, secretRef.Name)
+				}
+				secret, ok := obj.(*core_v1.Secret)
+				if !ok {
+					return nil, errors.Errorf("failure casting %v %q referenced from envFrom", gvk.Kind, secretRef.Name)
+				}
+
+				keys := make([]string, 0, len(secret.Data))
```
nilebox: This code partially duplicates `generateSecretParametersChecksumPayload` function used for ServiceInstance.
I think we can introduce a function like
```
func generateSecretChecksum(namespace string, name LocalObjectReference, key string /* optional */) ...
```

pkg/controller/bundlec/checksum.go
```diff
@@ -0,0 +1,25 @@
+package bundlec
+
+import (
+	"encoding/hex"
+
+	"golang.org/x/crypto/bcrypt"
+)
+
+func generateChecksum(data []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
nilebox: The linked code above is actually exactly why changing annotation causes creation of a new ReplicaSet - the hash label changes.

nilebox:
@ash2k @ychen-atlassian I think we can actually replace `serviceInstance.Spec.UpdateRequests++` with setting annotation with hash on `ServiceInstance` to make it more consistent with what we do in Deployment. Probably not in this PR though.

pkg/controller/bundlec/deployment.go
```diff
@@ -0,0 +1,154 @@
...
+				for k := range secret.Data {
+					keys = append(keys, k)
+				}
+				sort.Strings(keys)
+				for _, k := range keys {
+					secretData := secret.Data[k]
+					hash := sha256.Sum256(secretData)
```
ash2k: Why do we do double hashing? sha256 of each secret key, then this is all accumulated into a buffer and then the whole thing is bcrypted. Ok, this is how it is implemented right now....

Instead of accumulating into a buf we could use hasher itself as accumulator - like this:

```go
buf := sha256.New()
....
buf.Write(secretData)
...
```

pkg/controller/bundlec/deployment.go
```diff
@@ -0,0 +1,154 @@
...
+
+	for _, container := range template.Spec.Containers {
+		for _, envFrom := range container.EnvFrom {
+			if envFrom.SecretRef != nil {
+				secretRef := envFrom.SecretRef
+
+				gvk := core_v1.SchemeGroupVersion.WithKind("Secret")
+				obj, exists, err := st.store.Get(gvk, namespace, secretRef.Name)
+				if err != nil {
+					return nil, errors.Wrapf(err, "failure retrieving %v %q referenced from envFrom", gvk.Kind, secretRef.Name)
```
ash2k: Here and multiple places below: `%v` can be just `%s` for kind.

pkg/controller/bundlec/deployment.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	hasEnvFrom := false
+	for _, container := range deploymentSpec.Spec.Template.Spec.Containers {
+		if len(container.EnvFrom) != 0 {
+			hasEnvFrom = true
+			break
+		}
+	}
+	if !hasEnvFrom {
```
ash2k: This is not right - if there was a secret/configmap reference and it was removed, in the end the object will still have the annotation with the same value and will not be reprocessed by the deployment controller. But we want it to be reprocessed because the ref was removed, right?
We must change the value of the annotation here (or remove the annotation, but that is hard to do given current spec check implementation which merges annotations).

nilebox: @ash2k I agree that the annotation needs to be deleted, but I don't think leaving annotation means that Deployment won't be reprocessed. To remove envFrom reference you have to modify pod template spec, therefore replica set will be re-created with a normal logic, no annotation hack is required in that case.

ash2k: You are right, I didn't consider this case carefully enough, sorry.
I still think it's better to do something about the annotation - leaving it with previous value is misleading. Given that we cannot easily delete it... lets give it a different value. Empty string or perhaps hash of an empty slice to keep it uniform.

pkg/controller/bundlec/deployment.go
ychen-atlassian: It's probably not enough given LocalObjectReference usually looks at a concrete type and things probably need to generate checksums in different ways. I'll have a look at pulling this out into something that looks like ReadyChecker

pkg/controller/bundlec/deployment.go
ychen-atlassian: > Given that we cannot easily delete it

Hang on, so I can't just delete(..., myKey)?

pkg/controller/bundlec/deployment.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return unstructuredSpec, nil
+}
+
+func (st *resourceSyncTask) generateChecksum(template core_v1.PodTemplateSpec, namespace string) ([]byte, error) {
+	var buf bytes.Buffer
+
+	for _, container := range template.Spec.Containers {
+		for _, envFrom := range container.EnvFrom {
+			if envFrom.SecretRef != nil {
```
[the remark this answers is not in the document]
ychen-atlassian: I've opted to write something different. Will do another PR.

pkg/controller/bundlec/deployment.go
ash2k: No, because spec and actual object's annotations are merged. Yes, there is no way to remove an annotation via a Bundle - we need server-side apply. See
https://github.com/atlassian/smith/blob/master/pkg/speccheck/speccheck.go#L66
https://github.com/atlassian/smith/blob/master/pkg/speccheck/speccheck.go#L150-L163

pkg/controller/bundlec/deployment.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, errors.Wrap(err, "failed to marshal back into unstructured")
+	}
+
+	return unstructuredSpec, nil
+}
+
+func (st *resourceSyncTask) generateChecksum(template core_v1.PodTemplateSpec, namespace string) ([]byte, error) {
+	var buf bytes.Buffer
+
+	for _, container := range template.Spec.Containers {
```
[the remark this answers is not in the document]
ychen-atlassian: Done

pkg/controller/bundlec/hash.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	if !found && (optional == nil || !*optional) {
+		return errors.Errorf("not all keys %v found in configmap %q", filter, name)
+	}
+
+	return nil
+}
+
+// hashConfigMap hashes the sorted values in the configmap in sorted order
+// with a NUL as a separator character between and within pairs of key + value.
```
ash2k: This is awesome!

pkg/controller/bundlec/hash.go
```diff
@@ -0,0 +1,116 @@
...
+			return false, errors.WithStack(err)
+		}
+	}
+
+	return true, nil
+}
+
+// hashSecret hashes the sorted values in the secret in sorted order
+// with a NUL as a separator character between and within pairs of key + value.
+func hashSecret(secret *core_v1.Secret, h hash.Hash, filter sets.String) (bool, error) {
```
ash2k: Most of this code can/should be reused to support BinaryData of ConfigMap. Hash map of byte slices.

pkg/controller/bundlec/hash.go
```diff
@@ -0,0 +1,116 @@
...
+	if !found && (optional == nil || !*optional) {
+		return errors.Errorf("not all keys %v found in configmap %q", filter, name)
+	}
+
+	return nil
+}
+
+// hashConfigMap hashes the sorted values in the configmap in sorted order
+// with a NUL as a separator character between and within pairs of key + value.
+func hashConfigMap(configMap *core_v1.ConfigMap, h hash.Hash, filter sets.String) (bool, error) {
```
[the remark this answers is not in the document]
ash2k: You could also keep the hash type and ignore the errors. It is documented to never return errors. Less boilerplate. Or just use `utilruntime.Must()` as an error sink to make sure nothing weird happens unnoticed.

ychen-atlassian:
> @ash2k @ychen-atlassian I think we can actually replace `serviceInstance.Spec.UpdateRequests++` with setting annotation with hash on `ServiceInstance` to make it more consistent with what we do in Deployment. Probably not in this PR though.

I think we're already setting a hash with annotation on `ServiceInstance` 🤔. Maybe service catalog behaves differently with annotations?

pkg/controller/bundlec/hash.go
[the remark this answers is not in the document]
ychen-atlassian: I've opted to remain using hash.Hash since the point of this method is to pump things into a hash for later hashing, but I cast to io.Writer to enable the use of io.WriteString

pkg/controller/bundlec/hash.go
ychen-atlassian: I attempted to extract this but having the ConfigMap keys remain sorted even amongst binarydata (validation ensures none of the keys can clash) means there's not a whole lot of code here that can be deduped.

pkg/controller/bundlec/hash.go
```diff
@@ -0,0 +1,119 @@
...
+		return errors.Errorf("not all keys %v found in configmap %q", filter, name)
+	}
+
+	return nil
+}
+
+// hashConfigMap hashes the sorted values in the configmap in sorted order
+// with a NUL as a separator character between and within pairs of key + value.
+func hashConfigMap(configMap *core_v1.ConfigMap, h hash.Hash, filter sets.String) bool {
+	w := h.(io.Writer)
```
ash2k: This cast is not needed I think, or is it?

pkg/controller/bundlec/hash.go
```diff
@@ -0,0 +1,119 @@
...
+		w.Write([]byte{0})
+	}
+
+	return true
+}
+
+// hashSecret hashes the sorted values in the secret in sorted order
+// with a NUL as a separator character between and within pairs of key + value.
+func hashSecret(secret *core_v1.Secret, h hash.Hash, filter sets.String) bool {
+	w := h.(io.Writer)
```
ash2k: And this too

pkg/controller/bundlec/deployment.go
```diff
@@ -0,0 +1,115 @@
...
+	if v, ok := deploymentSpec.Spec.Template.ObjectMeta.Annotations[envRefHashAnnotation]; ok && v == disabled {
+		return spec, nil
+	}
+
+	bytes, err := st.generateHash(deploymentSpec.Spec.Template, namespace)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to generate checksum")
+	}
+
+	setDeploymentAnnotation(deploymentSpec, hex.EncodeToString(bytes))
```
ash2k: We used to bcrypt this hash to ensure there is no way to find a collision... but I'm not sure there is a chance to exploit this in any way. Also bcrypt is (by design) expensive cpu-wise, so maybe it is a good thing we remove it here. /cc @wryun @nilebox

nilebox: I don't think we should be concerned about collision probability here. Even if it happens once in a million years, we can just remove the annotation in that case 😛

nilebox: Ah, the concern is that someone decrypts a Secret via finding a hash key? That's unlikely given that data is not password, it would be silly to brute force or check vocabulary against JSON blob.

ychen-atlassian: Or we can add a salt? I can put this as another change if needed

ash2k: We have the salt there already - zero bytes and key names. Should be good enough.

PR #360, opened 2018-11-01
Simplify object spec pre-processing

ash2k:
No functional changes.

### DIFF
--- a/pkg/controller/bundlec/deployment.go
+++ b/pkg/controller/bundlec/deployment.go
@@ -10,7 +10,6 @@ import (
 	apps_v1 "k8s.io/api/apps/v1"
 	core_v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/controller/bundlec/deployment.go
```diff
@@ -19,17 +18,7 @@ const (
 )
 
 // works around https://github.com/kubernetes/kubernetes/issues/22368
```
nilebox: Comment should be moved/deleted?

nilebox: nvm, all good.

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -497,6 +479,20 @@ func (st *resourceSyncTask) evalSpec(res *smith_v1.Resource, actual runtime.Obje
 	}
 	obj.SetOwnerReferences(refs)
 
+	// Object GroupKind-specific munging
+	switch obj.GroupVersionKind().GroupKind() {
+	case schema.GroupKind{Group: sc_v1b1.GroupName, Kind: "ServiceInstance"}:
```
nilebox: nit: should Smith have constants for all supported Kinds?

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -497,6 +479,20 @@ func (st *resourceSyncTask) evalSpec(res *smith_v1.Resource, actual runtime.Obje
...
+		obj, err = st.processServiceInstance(obj, actual, st.bundle.Namespace)
```
nilebox: Could just return there (and for deployment as well). Fine if you prefer to explicitly check error.

pkg/controller/bundlec/resource_sync_task.go
ash2k: Yes, would be good to do this across the whole codebase.

pkg/controller/bundlec/resource_sync_task.go
ash2k: Yes, I prefer that. It follows the style.

PR #362, opened 2018-11-01
Only generate 1 owner ref per referred resource

ash2k:


pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -467,7 +467,13 @@ func (st *resourceSyncTask) evalSpec(res *smith_v1.Resource, actual runtime.Obje
 		Controller:         &trueRef,
 		BlockOwnerDeletion: &trueRef,
 	})
+	setRefs := make(map[smith_v1.ResourceName]struct{}, len(res.References))
 	for _, dep := range res.References {
+		if _, ok := setRefs[dep.Resource]; ok {
```
ychen-atlassian: So for each resource we generate an owner ref to the bundle. As well as this, we also generate an owner ref to any resources that reference it.

How come we only want to reference a single resource and not all of them in the bundle? I mean from a graph perspective it makes sense if we have edges to all other nodes that depend on us?

PR #363, opened 2018-11-01
Remove redundant namespace passing

ash2k:


ychen-atlassian:
Oops. I thought I fixed this 🤔

ash2k:
@ychen-atlassian nw, it is better to do this as a separate PR anyway. Otherwise it'd be too big.

PR #366, opened 2018-11-05
Move ServiceInstance and Deployment processing

ash2k:
This PR fixed the following issues:
- `LastAppliedReplicas` annotation is added by the "cleanup" code in speccheck. It should be added for objects that are to be created too. Otherwise specification re-check that compares spec vs what k8s has returned after create fails. This produces the kind of error in logs that is typical an indication of a bug in the system - this update result vs spec check was introduced to catch such bugs. And it is a bug indeed.
- Logic that belongs together is spread around several places. `Deployment` and `ServiceInstance` pre-processing is now each in a single file, not across packages.


cmd/smith/app/bundle_controller.go
```diff
@@ -194,6 +178,23 @@ func (c *BundleControllerConstructor) New(config *ctrl.Config, cctx *ctrl.Contex
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Spec check
+	checkTypes := []map[schema.GroupKind]speccheck.ObjectProcessor{builtin.MainKnownTypes}
+	if c.ServiceCatalogSupport {
+		checkTypes = append(checkTypes, builtin.ServiceCatalogKnownTypes)
+	}
+	specCheck := speccheck.New(multiStore, checkTypes...)
```
ash2k: Moved below because speccheck now needs the store.

it/deployment_ready_test.go
```diff
@@ -16,19 +14,13 @@ import (
...
+const (
+	deploymentResourceName smith_v1.ResourceName = "deployment-ready-test"
+)
 
 func constructBundle(t *testing.T, progressDeadlineSeconds int32, containerParams ...string) *smith_v1.Bundle {
-	// This is a hack for now until someone can help
-	// This is "set" in the deploymentCleanup but that is AFTER the object was created
-	// So this never really works since it should be set before the object was created
-	// in processResource "createOrUpdate"
-	const LastAppliedReplicasAnnotation string = smith.Domain + "/LastAppliedReplicas"
```
ash2k: Hack is not needed anymore! cc @roaanv

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -485,20 +484,6 @@ func (st *resourceSyncTask) evalSpec(res *smith_v1.Resource, actual runtime.Obje
 	}
 	obj.SetOwnerReferences(refs)
 
-	// Object GroupKind-specific munging
```
ash2k: This was moved into speccheck into corresponding files.

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -605,18 +590,23 @@ func (st *resourceSyncTask) createOrUpdate(spec *unstructured.Unstructured, actu
...
+		return st.createResource(resClient, spec)
+	default:
 		return st.updateResource(resClient, spec, actual)
 	}
-	st.logger.Info("Object not found, creating", ctrlLogz.ObjectGk(gvk.GroupKind()), ctrlLogz.Object(spec))
-	return st.createResource(resClient, spec)
 }
 
 func (st *resourceSyncTask) createResource(resClient dynamic.ResourceInterface, spec *unstructured.Unstructured) (actualRet *unstructured.Unstructured, retriableError bool, e error) {
+	spec, err := st.specCheck.BeforeCreate(st.logger, spec)
```
ash2k: Here we call speccheck instead of the removed switch above.

pkg/speccheck/builtin/process_deployment.go
```diff
@@ -1,27 +1,51 @@
...
+	var deploymentSpec apps_v1.Deployment
+	if err := util.ConvertType(appsV1Scheme, spec, &deploymentSpec); err != nil {
+		return nil, err
+	}
+
+	if deploymentSpec.Annotations == nil {
+		deploymentSpec.Annotations = make(map[string]string)
+	}
+	d.setLastAppliedReplicasAnnotation(ctx, &deploymentSpec, nil)
+	err := d.setConfigurationHashAnnotation(ctx, &deploymentSpec)
```
ash2k: Note how both methods are invoking these functions (see same file below). This achieves same behaviour for both objects that need to be updated and new objects that are created.

pkg/speccheck/builtin/process_deployment.go
```diff
@@ -33,46 +57,136 @@ func (deployment) ApplySpec(ctx *speccheck.Context, spec, actual *unstructured.U
...
-	if deploymentSpec.Annotations == nil {
-		deploymentSpec.Annotations = make(map[string]string)
+	return &deploymentSpec, nil
+}
+
+// setLastAppliedReplicasAnnotation updates replicas based on LastAppliedReplicas annotation and running config
+// to avoid conflicts with other controllers like HPA.
+// actual may be nil.
+func (deployment) setLastAppliedReplicasAnnotation(ctx *speccheck.Context, spec, actual *apps_v1.Deployment) {
+	if spec.Spec.Replicas == nil {
```
ash2k: If not set, we default to 1 and set the annotation accordingly.

pkg/speccheck/builtin/process_deployment_test.go
```diff
@@ -87,67 +79,65 @@ func TestAddsHashToDeploymentSpecForEnvFrom(t *testing.T) {
 
 	spec := runtimeToUnstructured(t, &deploymentSpec)
 
-	rst := resourceSyncTask{
```
ash2k: We now don't need to construct a bogus `resourceSyncTask`. Just test stuff directly.

pkg/speccheck/builtin/process_secret.go
```diff
@@ -11,6 +11,10 @@ import (
 type secret struct {
 }
 
+func (secret) BeforeCreate(ctx *speccheck.Context, spec *unstructured.Unstructured) (runtime.Object /*updatedSpec*/, error) {
```
ash2k: All processors except for the ServiceInstance and Deployment ones just return the spec verbatim.

pkg/speccheck/builtin/process_service_instance.go
```diff
@@ -1,17 +1,39 @@
...
 type serviceInstance struct {
 }
 
-func (serviceInstance) ApplySpec(ctx *speccheck.Context, spec, actual *unstructured.Unstructured) (runtime.Object, error) {
+func (s serviceInstance) BeforeCreate(ctx *speccheck.Context, spec *unstructured.Unstructured) (runtime.Object /*updatedSpec*/, error) {
+	var instanceSpec sc_v1b1.ServiceInstance
+	if err := util.ConvertType(scV1B1Scheme, spec, &instanceSpec); err != nil {
+		return nil, err
+	}
+	if err := s.setSecretParametersChecksumAnnotation(ctx, &instanceSpec, nil); err != nil {
```
ash2k: Note how this method is invoked in both `BeforeCreate` and `ApplySpec`.

pkg/speccheck/builtin/process_service_instance.go
```diff
@@ -42,3 +67,74 @@ func (serviceInstance) ApplySpec(ctx *speccheck.Context, spec, actual *unstructu
...
+//
+// The annotation is only ever added to the spec object not the actual object. The modified spec is treated as
+// if the user manually set the annotation.
+//
+// If the a user actually did set the annotation manually there are two cases.
+// 1. if there are no parametersFrom referenced secrets we leave the spec untouched
+// 2. if there are referenced secrets we proceed to check and generate a new value for the annotation
+//
+// This can be disabled by adding the annotation with the value 'disabled'
+func (s serviceInstance) setSecretParametersChecksumAnnotation(ctx *speccheck.Context, spec, actual *sc_v1b1.ServiceInstance) error {
```
ash2k: This logic is mostly left untouched, however, we now always set the annotation regardless of whether there are secret references or not. (need to update the comment above!)

pkg/speccheck/builtin/process_service_instance.go
```diff
@@ -1,17 +1,39 @@
...
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	SecretParametersChecksumAnnotation = smith.Domain + "/secretParametersChecksum"
+	Disabled                           = "disabled"
 )
 
 type serviceInstance struct {
```
ash2k: All the static functions here and for Deployment are now methods to avoid polluting package's name space. Re-useable ones for hashing are in the `speccheck` package, everything else is not re-useable anyway.

pkg/speccheck/builtin/process_service_instance_test.go
```diff
@@ -0,0 +1,368 @@
+package builtin
```
ash2k: These tests are mostly the same. I've modified some of them to ensure the annotation is always set (see previous comment).

pkg/speccheck/speccheck.go
```diff
@@ -28,10 +29,28 @@ func New(kts ...map[schema.GroupKind]ObjectProcessor) *SpecCheck {
...
 	}
 	return &SpecCheck{
+		Store:      store,
 		KnownTypes: kt,
 	}
 }
 
+// BeforeCreate pre-processes object specification and returns an updated version.
+func (sc *SpecCheck) BeforeCreate(logger *zap.Logger, spec *unstructured.Unstructured) (*unstructured.Unstructured /*updatedSpec*/, error) {
+	processor, ok := sc.KnownTypes[spec.GroupVersionKind().GroupKind()]
```
ash2k: The removed `switch` from the `resource_sync_task.go` is this lookup now essentially (and the same lookup in `compareActualVsSpec()` below).

pkg/speccheck/testing/stuff_for_tests.go
```diff
@@ -0,0 +1,20 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type FakeStore struct {
+	Namespace string
+	Responses map[string]runtime.Object
+}
+
+func (f FakeStore) Get(gvk schema.GroupVersionKind, namespace, name string) (runtime.Object, bool /*exists*/, error) {
+	if f.Namespace != namespace {
```
ash2k: Currently `Namespace` field is not being used anywhere and namespace is not set for objects in tests. I'm planning to send a follow up PR to set it to have tests that verify that namespace is being passed correctly.

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -605,18 +590,23 @@ func (st *resourceSyncTask) createOrUpdate(spec *unstructured.Unstructured, actu
 	gvk := spec.GroupVersionKind()
 	resClient, err := st.smartClient.ForGVK(gvk, st.bundle.Namespace)
 	if err != nil {
-		return nil, false, errors.Wrapf(err, "failed to get the client for %q", gvk)
+		return nil, false, errors.Wrapf(err, "failed to get the client for %s", gvk)
 	}
-	if actual != nil {
-		st.logger.Info("Object found, checking spec", ctrlLogz.ObjectGk(gvk.GroupKind()), ctrlLogz.Object(spec))
+	switch actual {
```
amckague: Just use if/else?

pkg/speccheck/builtin/process_deployment.go
```diff
@@ -33,46 +57,136 @@ func (deployment) ApplySpec(ctx *speccheck.Context, spec, actual *unstructured.U
...
-	return util.RuntimeToUnstructured(&deploymentSpec)
+	if spec.Spec.Template.Annotations == nil {
+		spec.Spec.Template.Annotations = make(map[string]string, 1)
+	}
+	spec.Spec.Template.Annotations[EnvRefHashAnnotation] = hex.EncodeToString(hashBytes)
+	return nil
+}
+
+func (d deployment) generateHash(ctx *speccheck.Context, spec *apps_v1.Deployment) ([]byte, error) {
+	hasher := sha256.New()
```
amckague: Are you hashing over secret data at all?

PR #368, opened 2018-11-12
Fixed handling of empty namespace

ychen-atlassian:


pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -483,6 +483,13 @@ func (st *resourceSyncTask) evalSpec(res *smith_v1.Resource, actual runtime.Obje
 		})
 	}
 	obj.SetOwnerReferences(refs)
+	switch obj.GetNamespace() {
+	case "":
+		obj.SetNamespace(st.bundle.Namespace)
+	case st.bundle.Namespace:
+	default:
+		return nil, errors.Errorf("namespace of resource %q was %q which is different from the bundle namespace %q", res.Name, obj.GetNamespace(), st.bundle.Namespace)
```
ash2k: No need to mention resource name - error message is attached to a particular resource and is logged with resource in the logger field. This would also be consistent with how we do it in all other places.

PR #370, opened 2018-11-14
Reduce errors from external ServiceInstance failures

ychen-atlassian:
This follows on changes from my previous attempts to reduce errors.

1. Ready Checker should not be returning both the object status and the error, because we already capture whether or not the error is a "user error". Errors just get shoved into the Result object
2. This means we don't need to deal with weird error special case, and I've bumped down the logging in specific cases
3. Also copy all the messages and stuff
4. Also reduce log to Debug for many other lines

### DIFF
--- a/pkg/controller/bundlec/bundle_sync_task.go
+++ b/pkg/controller/bundlec/bundle_sync_task.go
@@ -564,8 +564,10 @@ func (st *bundleSyncTask) resourceConditions(res smith_v1.Resource) (
 			blockedCond.Message = fmt.Sprintf("Not ready: %q", resStatus.dependencies)
 		case resourceStatusInProgress:
 			inProgressCond.Status = cond_v1.ConditionTrue
+			inProgressCond.Message = resStatus.message
 		case resourceStatusReady:
 			readyCond.Status = cond_v1.ConditionTrue
+			readyCond.Message = resStatus.message
 		case resourceStatusError:
 			errorCond.Status = cond_v1.ConditionTrue
 			errorCond.Message = resStatus.err.Error()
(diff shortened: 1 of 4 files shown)


pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -99,15 +99,27 @@ func (st *bundleSyncTask) processNormal() (retriableError bool, e error) {
...
+				return resErr.isRetriableError, resErr.err
+			}
+			if !resErr.isExternalError {
+				logger.Error("Done processing resource with internal error",
+					zap.Bool("ready", resInfo.isReady()),
+					zap.Error(resErr.err),
+					zap.Bool("retriable", resErr.isRetriableError))
+			} else {
+				// Log at info level - external errors are expected (e.g. if the
+				// user puts something invalid in the spec)
```
ychen-atlassian: Also useful to log at Info level because it's useful info

pkg/statuschecker/ready_checker.go
```diff
@@ -88,12 +88,14 @@ func New(store CRDStore, kts ...map[schema.GroupKind]ObjectStatusChecker) (*Chec
...
 
-func (c *Checker) CheckStatus(obj *unstructured.Unstructured) (ObjectStatusResult, error) {
+func (c *Checker) CheckStatus(obj *unstructured.Unstructured) ObjectStatusResult {
 	gvk := obj.GroupVersionKind()
 	gk := gvk.GroupKind()
 
 	if gk.Kind == "" || gvk.Version == "" { // Group can be empty e.g. built-in objects like ConfigMap
-		return ObjectStatusUnknown{}, errors.Errorf("object has empty kind/version: %s", gvk)
+		return ObjectStatusError{
+			Error: errors.Errorf("object has empty kind/version: %s", gvk),
```
ash2k: This is an internal error

ash2k: Hm, I'm not sure it always is - I don't remember if we validate objects in the bundle. But I'm sure I added this check to catch internal bugs and it helped quite a few times. Can you please check if we validate that objects in the Bundle have these fields set correctly.

ychen-atlassian: Yeah it's marked as internal - the ObjectStatusError also has an "IsExternalError" field

PR #372, opened 2018-11-15
Reduce errors in spec processing

ychen-atlassian:
Another PR to change the errors from spec processing into using resourceStatus instead.


pkg/controller/bundlec/spec_processor.go
```diff
@@ -164,7 +164,7 @@ func (sp *specProcessor) ProcessString(value string, path ...string) (interface{
 	return reference, nil
 }
 
-func resolveReference(resInfos map[smith_v1.ResourceName]*resourceInfo, reference smith_v1.Reference) (interface{}, error) {
+func jsonPathReferenceResolver(resInfos map[smith_v1.ResourceName]*resourceInfo, reference smith_v1.Reference) (interface{}, error) {
```
ychen-atlassian: Unrelated change - I renamed this because in a function above, we have a function parameter also named "resolveReference", which is SUPER confusing when glancing at the code since it is not immediately obvious which function is being called

PR #381, opened 2018-11-22
multirun does not work with rules_docker

ash2k:
As we have discovered today, multirun does not work properly with cross-compilation for a complicated reason.


ash2k:
Documented the issue here https://groups.google.com/d/msg/bazel-discuss/BoK0lojLWUU/Gu7ZRPv3AQAJ

PR #382, opened 2018-11-23
Support setting configuration file path

ash2k:


Makefile
```diff
@@ -1,6 +1,8 @@
-OS = $$(uname -s | tr A-Z a-z)
-BINARY_PREFIX_DIRECTORY=$(OS)_amd64_stripped
-KUBE_CONTEXT ?= minikube
+OS := $(shell uname -s | tr A-Z a-z)
+BINARY_PREFIX_DIRECTORY := $(OS)_amd64_stripped
+KUBECONTEXT ?= kubernetes-admin@kind-1
+KUBECONFIG ?= $(shell kind get kubeconfig-path)
+#KUBECONFIG ?= $$HOME/.kube/config
```
nilebox: Remove the comment and add a text comment instead referring to `kind`?
`KUBECONTEXT` is also kind-specific...

Makefile
```diff
@@ -1,6 +1,8 @@
-OS = $$(uname -s | tr A-Z a-z)
-BINARY_PREFIX_DIRECTORY=$(OS)_amd64_stripped
-KUBE_CONTEXT ?= minikube
+OS := $(shell uname -s | tr A-Z a-z)
+BINARY_PREFIX_DIRECTORY := $(OS)_amd64_stripped
+KUBECONTEXT ?= kubernetes-admin@kind-1
```
benbarclay: It's probably worthwhile to give kind a `--name` argument so that you have an isolated cluster for the project. This creates a completely different cluster + kubeconfig file.

The `@kind-1` comes from the default cluster name `1`

This is only really a concern if you're creating a cluster in here (which it appears you are not) but might be worth keeping in mind if you do end up creating one for CI or something down the track.

PR #391, opened 2018-12-04
Reduce error rates from Bad Request on create/apply of object

ychen-atlassian:


pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -151,6 +151,15 @@ func (st *resourceSyncTask) processResource(res *smith_v1.Resource) resourceInfo
 	// Create or update resource
 	resUpdated, retriable, err := st.createOrUpdate(spec, actual)
 	if err != nil {
+		if api_errors.IsBadRequest(err) {
```
ychen-atlassian: I'm not sure if this is actually the cause. Creating some objects results in a schema validation failure sometimes, e.g. when deployment port is longer than 15 characters. I'm assuming this comes in as a Bad Request.

ychen-atlassian:
```
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Deployment.apps \"ups-broker-ups-broker\" is invalid: spec.template.spec.containers[0].livenessProbe.tcpSocket.port: Invalid value: \"8080dododododododododododododoodododododododod\": must be no more than 15 characters","reason":"Invalid","details":{"name":"ups-broker-ups-broker","group":"apps","kind":"Deployment","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: \"8080dododododododododododododoodododododododod\": must be no more than 15 characters","field":"spec.template.spec.containers[0].livenessProbe.tcpSocket.port"}]},"code":422}
```

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -151,6 +151,15 @@ func (st *resourceSyncTask) processResource(res *smith_v1.Resource) resourceInfo
 	// Create or update resource
 	resUpdated, retriable, err := st.createOrUpdate(spec, actual)
 	if err != nil {
+		if api_errors.IsInvalid(err) {
```
ash2k: We may wrap the error later and that will break this logic. Would be good to add a test for this case and change the code to extract the cause like this: `if api_errors.IsInvalid(errors.Cause(err)) {`.

ash2k:
Rebased + added error unwrapping.

PR #395, opened 2019-01-04
Split resource status into two structs for reuse

ash2k:


pkg/apis/smith/v1/zz_generated.deepcopy.go
```diff
@@ -204,6 +204,23 @@ func (in *ResourceSpec) DeepCopy() *ResourceSpec {
 
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
 func (in *ResourceStatus) DeepCopyInto(out *ResourceStatus) {
+	*out = *in
```
benbarclay: Wait what?

ash2k: This is the standard way that the generator uses to copy all fields that don't need to be deep copied. E.g. strings, numbers, bool, plain structs - anything that is not a slice/map/etc basically.

Issue #289, opened 2018-05-28
Validating wek hook for synchronous validation


ash2k:
Expose a validating admission hook endpoint to enable synchronous Bundle validation.

ash2k:
Perhaps it should be a different binary in the same docker image deployed as a separate Deployment to avoid coupling - if smith is crashing/etc it should not influence the webhook.

ash2k:
A validation step: prohibit Bundles with resources with annotation `DeletionTimestampAnnotation `. See https://github.com/atlassian/smith/pull/396#discussion_r246263249

PR #396, opened 2019-01-04
Initial support for soft delete

nilebox:
Initial code for soft delete support.

### DIFF
--- a/api.go
+++ b/api.go
@@ -12,6 +12,9 @@ const (
 	EventAnnotationResourceName = Domain + "/ResourceName"
 	EventAnnotationReason       = Domain + "/Reason"
 
+	DeletionDelayAnnotation     = Domain + "/deleteDelay"
+	DeletionTimestampAnnotation = Domain + "/deletionTimestamp"
+
 	EventReasonResourceInProgress = "ResourceInProgress"
 	EventReasonResourceReady      = "ResourceReady"
 	EventReasonResourceError      = "ResourceError"
--- a/api.go
+++ b/api.go
@@ -12,7 +12,7 @@ const (
 	EventAnnotationResourceName = Domain + "/ResourceName"
 	EventAnnotationReason       = Domain + "/Reason"
 
-	DeletionDelayAnnotation     = Domain + "/deleteDelay"
+	DeletionDelayAnnotation     = Domain + "/deletionDelay"
 	DeletionTimestampAnnotation = Domain + "/deletionTimestamp"
 
 	EventReasonResourceInProgress = "ResourceInProgress"
--- a/pkg/controller/bundlec/bundle_sync_task.go
+++ b/pkg/controller/bundlec/bundle_sync_task.go
@@ -339,7 +339,7 @@ func (st *bundleSyncTask) preDelete(logger *zap.Logger, name string, obj runtime
 	deletionTimestampAnnotation, ok := annotations[smith.DeletionTimestampAnnotation]
 	if !ok {
 		// Mark object with deletionTimestamp annotation to start the countdown
-		annotations[smith.DeletionTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
+		annotations[smith.DeletionTimestampAnnotation] = timeToString(time.Now())
 		m.SetAnnotations(annotations)
 
 		unstr, err := util.RuntimeToUnstructured(obj)
@@ -360,7 +360,7 @@ func (st *bundleSyncTask) preDelete(logger *zap.Logger, name string, obj runtime
 		return false, false, nil
 	}
 	// Check if deletion delay has expired
-	deletionTimestamp, err := time.Parse(time.RFC3339, deletionTimestampAnnotation)
+	deletionTimestamp, err := timeFromString(deletionTimestampAnnotation)
 	if err != nil {
 		return false, false, errors.Wrap(err, "failed to unmarshal deletion timestamp")
 	}
@@ -787,4 +787,12 @@ func sortBundle(bundle *smith_v1.Bundle) (*graph.Graph, []graph.V, error) {
 	}
 
 	return g, sorted, nil
+}
+
+func timeToString(t time.Time) string {
+	return t.UTC().Format(time.RFC3339)
+}
+
+func timeFromString(s string) (time.Time, error) {
+	return time.Parse(time.RFC3339, s)
 }
--- a/pkg/controller/bundlec/bundle_sync_task.go
+++ b/pkg/controller/bundlec/bundle_sync_task.go
@@ -429,13 +429,13 @@ func (st *bundleSyncTask) handleNormalStatusUpdate(retriable bool, processErr er
 	// Construct resource conditions and check if there were any resource errors
 	resourceStatuses := make([]smith_v1.ResourceStatus, 0, len(st.processedResources))
 	var failedResources []smith_v1.ResourceName
-	retriableResourceErr := true
+	retriableResourceErr := false
 	for _, res := range st.bundle.Spec.Resources { // Deterministic iteration order
 		blockedCond, inProgressCond, readyCond, errorCond := st.resourceConditions(res)
 
 		if errorCond.Status == cond_v1.ConditionTrue {
 			failedResources = append(failedResources, res.Name)
-			retriableResourceErr = retriableResourceErr && errorCond.Reason == smith_v1.ResourceReasonRetriableError // Must not continue if at least one error is not retriable
+			retriableResourceErr = retriableResourceErr || errorCond.Reason == smith_v1.ResourceReasonRetriableError // Must continue if at least one error is retriable
 		}
 
 		bundleStatusUpdated = st.checkResourceConditionNeedsUpdate(res.Name, &blockedCond) || bundleStatusUpdated
--- a/pkg/controller/bundlec/bundle_sync_task.go
+++ b/pkg/controller/bundlec/bundle_sync_task.go
@@ -172,7 +172,7 @@ func (st *bundleSyncTask) deleteAllResources() (retriableError bool, e error) {
 	st.objectsToDelete = make(map[objectRef]runtime.Object, len(objs))
 
 	var firstErr error
-	retriable := true
+	retriable := false
 	policy := meta_v1.DeletePropagationForeground
 	for _, obj := range objs {
 		m := obj.(meta_v1.Object)
@@ -195,7 +195,6 @@ func (st *bundleSyncTask) deleteAllResources() (retriableError bool, e error) {
 		resClient, err := st.smartClient.ForGVK(gvk, st.bundle.Namespace)
 		if err != nil {
 			if firstErr == nil {
-				retriable = false
 				firstErr = err
 			} else {
 				logger.Error("Failed to get client for object", zap.Error(err))
@@ -214,6 +213,7 @@ func (st *bundleSyncTask) deleteAllResources() (retriableError bool, e error) {
 			// conflict means it has been deleted and re-created (UID does not match)
 			if firstErr == nil {
 				firstErr = err
+				retriable = true // could be some temporary network issue
 			} else {
 				logger.Warn("Failed to delete object", zap.Error(err))
 			}
@@ -278,7 +278,6 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
 		resClient, err := st.smartClient.ForGVK(ref.GroupVersionKind, st.bundle.Namespace)
 		if err != nil {
 			if firstErr == nil {
-				retriable = false
 				firstErr = err
 			} else {
 				logger.Error("Failed to get client for object", zap.Error(err))
@@ -288,9 +287,11 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
 
 		readyToDelete, retriableErr, err := st.preDelete(logger, ref.Name, obj, resClient)
 		if err != nil {
+			if retriableErr {
+				retriable = true
+			}
 			if firstErr == nil {
 				firstErr = err
-				retriable = retriableErr
 			} else {
 				logger.Error("Failed to execute pre-delete for object", zap.Error(err))
 			}
--- a/pkg/controller/bundlec/bundle_sync_task.go
+++ b/pkg/controller/bundlec/bundle_sync_task.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/dynamic"
 )
 
 type bundleSyncTask struct {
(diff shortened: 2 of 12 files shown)


api.go
```diff
@@ -12,6 +12,9 @@ const (
 	EventAnnotationResourceName = Domain + "/ResourceName"
 	EventAnnotationReason       = Domain + "/Reason"
 
+	DeletionDelayAnnotation     = Domain + "/deleteDelay"
```
nilebox: TODO: rename annotation to `/deletionDelay`

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -25,6 +25,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/tools/record"
+	"time"
```
nilebox: TODO: fix import ordering

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,55 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
 			continue
 		}
 
+		deletionDelayAnnotation, ok := m.GetAnnotations()[smith.DeletionDelayAnnotation]
+		if ok {
+			// Trigger the "deletion delay" logic
+			deletionDelay, err := time.ParseDuration(deletionDelayAnnotation)
+			if err != nil {
+				return false, errors.Wrap(err, "failed to parse deletion delay duration")
```
ash2k: This should not stop the loop. Just log and continue to the next object. Also may emit an event on the bundle or on the object.

ash2k: Include the annotation name and value that failed to parse? Maybe that is too verbose... But it shouldn't happen often or at all so verbose is ok =)

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,55 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+			}
+			deletionTimestampAnnotation, ok := m.GetAnnotations()[smith.DeletionTimestampAnnotation]
+			if !ok {
+				// Mark object with deletionTimestamp annotation to start the countdown
+				annotations := m.GetAnnotations()
+				deletionTimestampBytes, err := meta_v1.Now().MarshalJSON()
```
ash2k: This is weird. Why not pick a format (take some pre-existing RFC-defined format from stdlib) and do `fmt.Sprintf()` on it? Which format is used should be defined in the spec document.

ash2k: This approach will also spare you from the error below

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,55 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+				}
+				annotations[smith.DeletionTimestampAnnotation] = string(deletionTimestampBytes)
+				m.SetAnnotations(annotations)
+
+				unstr, err := util.RuntimeToUnstructured(obj)
+				if err != nil {
+					return false, errors.Wrap(err, "failed to convert runtime object to unstructured")
+				}
+				_, err = resClient.Update(unstr, meta_v1.UpdateOptions{})
+				if err != nil {
```
ash2k: If this is a conflict - exit the loop early because bundle will be re-enqueued.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,55 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+					if firstErr == nil {
+						firstErr = err
+					} else {
+						logger.Warn("Failed to update object", zap.Error(err))
+					}
+					continue
+				}
+				// The object will eventually be reprocessed for the check for delay expiration
```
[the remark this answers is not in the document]
ash2k: I think it is fine at least for now/first iteration. Later we can have an index on all informers to get all objects marked with that annotation and re-check them regularly (e.g. every 1 minute) or be even smarter and sleep until the next object should be deleted but no longer than some threshold (too smart probably).

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,68 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+					logger.Warn("failed to parse deletion delay duration", zap.Error(err))
+				}
+				continue
+			}
+			deletionTimestampAnnotation, ok := m.GetAnnotations()[smith.DeletionTimestampAnnotation]
+			if !ok {
+				// Mark object with deletionTimestamp annotation to start the countdown
+				annotations := m.GetAnnotations()
+				annotations[smith.DeletionTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
+				m.SetAnnotations(annotations)
```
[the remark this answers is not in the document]
nilebox: @roaanv  only if `GetAnnotations()` returns the internal map rather than a copy of a map, which is implementation details.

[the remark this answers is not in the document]
nilebox: Just looked up:
- `ObjectMeta` indeed just returns internal map: https://github.com/kubernetes/apimachinery/blob/849b284f3b756a3fbb5084b09d2718dd41373068/pkg/apis/meta/v1/meta.go#L159
- but `Unstructured` returns a new map every time: https://github.com/kubernetes/apimachinery/blob/849b284f3b756a3fbb5084b09d2718dd41373068/pkg/apis/meta/v1/unstructured/unstructured.go#L377-L380

so no, I can't rely on that

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,68 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+
+				unstr, err := util.RuntimeToUnstructured(obj)
```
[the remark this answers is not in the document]
nilebox: @roaanv only if we cast `m` back to `runtime.Object`... but we got `m` by casting an `obj runtime.Object` to `meta_v1.Object`.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,68 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+						return retriable, firstErr
+					}
+					continue
+				}
+				// The object will eventually be reprocessed for the check for delay expiration
+				continue
+			}
+			// Check if deletion delay has expired
+			deletionTimestamp, err := time.Parse(time.RFC3339, deletionTimestampAnnotation)
+			if err != nil {
```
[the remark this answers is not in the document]
nilebox: Good catch!

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,68 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+				if firstErr == nil {
+					retriable = false
+					firstErr = errors.Wrap(err, "failed to parse deletion delay duration")
+				} else {
+					logger.Warn("failed to parse deletion delay duration", zap.Error(err))
+				}
+				continue
+			}
+			deletionTimestampAnnotation, ok := m.GetAnnotations()[smith.DeletionTimestampAnnotation]
+			if !ok {
```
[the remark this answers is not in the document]
nilebox: yep, that's what I planned to do.

pkg/controller/bundlec/bundle_sync_task.go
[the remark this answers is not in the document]
roaanv: :(

pkg/controller/bundlec/bundle_sync_task.go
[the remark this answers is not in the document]
roaanv: :(

pkg/controller/bundlec/bundle_sync_task.go
nilebox: moving code to a separate function helped.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +285,68 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+				// Mark object with deletionTimestamp annotation to start the countdown
+				annotations := m.GetAnnotations()
+				annotations[smith.DeletionTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
```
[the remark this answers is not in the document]
nilebox: you mean some kind of `marshalTime` and `unmarshalTime` functions? can declare just in this file, then we can move these functions to some utility package if we need to reuse them.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -303,6 +322,58 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
 	return retriable, firstErr
 }
 
+func (st *bundleSyncTask) preDelete(logger *zap.Logger, name string, obj runtime.Object, m meta_v1.Object, resClient dynamic.ResourceInterface) (bool /* readyToDelete */, bool /* retriableError */, error) {
```
ash2k: Don't pass both `obj` and `m`. Just `obj` is enough.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -303,6 +322,58 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+		}
+		_, err = resClient.Update(unstr, meta_v1.UpdateOptions{})
+		if err != nil {
+			if api_errors.IsConflict(err) {
+				// Escape the processing loop, the conflict means that
+				// we are processing a stale object revision,
+				// and a newer revision will be re-processed soon
+				return false, false, err
+			}
+			return false, true, err
```
ash2k: Ignore error (continue iteration) if it was `not found`

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -303,6 +322,58 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+		}
+		// The object will eventually be reprocessed for the check for delay expiration
+		return false, false, nil
+	}
+	// Check if deletion delay has expired
+	deletionTimestamp, err := timeFromString(deletionTimestampAnnotation)
+	if err != nil {
+		return false, false, errors.Wrap(err, "failed to unmarshal deletion timestamp")
+	}
+	if time.Now().UTC().Before(deletionTimestamp.Add(deletionDelay)) {
```
ash2k: No need to convert to UTC. Time zone does not matter when comparing time objects because absolute timestamp values are compared.

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -303,6 +322,58 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
...
+		// Skip deletion of resource until the delay expires
+		return false, false, nil
+	}
+	// All checks passed -> deletion delay has expired,
+	// and we can proceed with deletion
+	logger.Sugar().Debugf("Deletion delay has expired, proceeding: name=%q, delay=%v, timestamp=%v", name, deletionDelay, deletionTimestamp)
```
ash2k: Object name should be attached to the logger already, isn't it?

pkg/controller/bundlec_test/delay_start_delete_removed_object_test.go
```diff
@@ -0,0 +1,51 @@
...
+				Namespace:  testNamespace,
+				UID:        bundle1uid,
+				Finalizers: []string{bundlec.FinalizerDeleteResources},
+			},
+		},
+		expectedActions: sets.NewString("PUT=/api/v1/namespaces/" + testNamespace + "/configmaps/" + mapNeedsAnUpdate),
+		appName:         testAppName,
+		namespace:       meta_v1.NamespaceAll,
+		testHandler: fakeActionHandler{
+			response: map[path]fakeResponse{
```
nilebox: I was also going to test that the request body contains a `deletionTimestamp` annotation, for which I need to intercept the request body, and after reading it I would have to re-create some implementation of the `io.ReadCloser`, so that the next handler in the chain will be able to read it.
I've found an `httputil.DumpRequest` that does that (https://rominirani.com/golang-tip-capturing-http-client-requests-incoming-and-outgoing-ef7fcdf87113), but it returns a raw formatted string instead of a structured result :( We can still the implementation code from this function if we want, but for now I decided not to bother...

pkg/controller/bundlec/bundle_sync_task.go
nilebox: Yep, you're right

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -717,6 +718,15 @@ func (st *resourceSyncTask) updateResource(resClient dynamic.ResourceInterface,
 	if err != nil {
 		return nil, false, errors.Wrap(err, "specification check failed")
 	}
+
+	// Delete the DeletionTimestamp annotation if it is present
+	annotations := updated.GetAnnotations()
```
nilebox: @ash2k is this the correct place to put the annotation deletion code?

pkg/controller/bundlec_test/delay_proceed_delete_removed_object_test.go
```diff
@@ -0,0 +1,57 @@
...
+// Should proceed with deletion of the controlled object
+// after the deletion delay has expired
+func TestDelayDeleteProceed(t *testing.T) {
+	t.Parallel()
+	m1 := configMapNeedsUpdate()
+	if m1.Annotations == nil {
+		m1.Annotations = make(map[string]string)
+	}
+	m1.Annotations["smith.atlassian.com/deletionDelay"] = "1h"
+	m1.Annotations["smith.atlassian.com/deletionTimestamp"] = time.Now().Add(-2 * time.Hour).UTC().Format(time.RFC3339)
```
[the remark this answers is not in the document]
nilebox: to make sure we get via code what we expect (not so much about helper functions, more about annotation names)

pkg/controller/bundlec/resource_sync_task.go
ash2k: Made a note here https://github.com/atlassian/smith/issues/289#issuecomment-452577218

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -282,6 +286,21 @@ func (st *bundleSyncTask) deleteRemovedResources() (retriableError bool, e error
 			continue
 		}
 
+		readyToDelete, retriableErr, err := st.preDelete(logger, ref.Name, obj, resClient)
+		if err != nil {
+			if firstErr == nil {
+				firstErr = err
+				retriable = retriableErr
+			} else {
+				logger.Error("Failed to execute pre-delete for object", zap.Error(err))
```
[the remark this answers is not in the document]
nilebox: @ash2k we initialize `retriable` with true (i.e. by default), and reset it if we the first error is non-retriable... not sure if that was your initial intent, but that's how the code is written.

pkg/controller/bundlec/resource_sync_task.go
nilebox: @ash2k actually I would need to remove the annotation from `actual`, rather than from `spec` (AFAIK the "merge" logic will combine annotations from `actual` and `spec` => will keep the annotation in the result), and removing annotation from `actual` seems hacky to me (we shouldn't mutate `actual` before comparison?).
Mutating `updated` is fine (the merge result, "desired state"), on the other hand.
So I think I would prefer to keep the code unless I misunderstood something

pkg/controller/bundlec/resource_sync_task.go
```diff
@@ -551,6 +570,22 @@ func (st *resourceSyncTask) evalSpec(res *smith_v1.Resource, actual runtime.Obje
...
 
+// validateSpec enforces constraints on the desires object spec
+// e.g. prohibits Smith-managed annotations
+func (st *resourceSyncTask) validateSpec(spec *unstructured.Unstructured) resourceStatus {
+	annotations := spec.GetAnnotations()
+	if len(annotations) > 0 {
+		for key := range prohibitedAnnotations {
+			if _, ok := annotations[key]; ok {
+				return resourceStatusError{
+					err: errors.Errorf("annotation %q cannot be set by the user", key),
```
nilebox: TODO: add a unit test checking this actually works

pkg/controller/bundlec/bundle_sync_task.go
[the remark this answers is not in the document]
ash2k: I'm suggesting to change this to return `true` if at least one of the encountered errors is retriable.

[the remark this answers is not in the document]
nilebox: Agreed, that makes sense. I will rewrite the code a bit then.

nilebox:
The code is done, I just need to write a documentation for this feature.

PR #397, opened 2019-01-05
Enqueue Bundle on ConfigMap/Secret event

### LINKED ISSUE #394: Reprocess Bundles with Deployments when referred Secret/ConfigMaps change
Currently we don't have a reverse index from `ConfigMap`/`Secret` objects to the `Bundles` which have `Deployments` that use those objects. So when `ConfigMap`/`Secret` objects change, the `Bundle` that contain affected `Deployment`(s) are not re-enqueued.

ash2k:
If there is a Deployment in a Bundle that uses ConfigMap/Secret and this object is updated, Bundle is enqueued for processing.
Fixes #394.

Not sure how to test this. Will think more.


pkg/controller/bundlec/controller.go
```diff
@@ -125,3 +194,77 @@ func (c *controllerIndexAdapter) ControllerByObject(gk schema.GroupKind, namespa
...
+
+func configMapNamespaceNameIndexKeysForContainers(namespace string, containers []core_v1.Container) []string {
+	var indexKeys []string
+	for _, container := range containers {
+		for _, envFrom := range container.EnvFrom {
+			configMapRef := envFrom.ConfigMapRef
+			if configMapRef == nil {
+				continue
+			}
+			indexKeys = append(indexKeys, bySecretNamespaceNameIndexKey(namespace, configMapRef.Name))
```
Burr: this should be `byConfigMapNamespaceNameIndexKey`, right?

same with the statement below.

ash2k: Doh, I checked it like 3 times and still got it wrong =) Thanks! Copy-pasting is bad.

pkg/controller/bundlec/controller.go
```diff
@@ -125,3 +194,77 @@ func (c *controllerIndexAdapter) ControllerByObject(gk schema.GroupKind, namespa
...
+				continue
+			}
+			indexKeys = append(indexKeys, byConfigMapNamespaceNameIndexKey(namespace, configMapKeyRef.Name))
+		}
+	}
+	return indexKeys
+}
+
+func byConfigMapNamespaceNameIndexKey(configMapNamespace, configMapName string) string {
+	return configMapNamespace + "/" + configMapName
```
benbarclay: Any specific reason you use this style over `fmt.Sprintf`?

ash2k: It is not important in this case of course, but I think concatenation is faster - don't need to parse the formatting string. Also this code does not depend on the `fmt` package that way (also not important).

ash2k:
I want to have some tests here, so I'm not merging it straight away.

pkg/controller/bundlec/controller.go
```diff
@@ -110,6 +148,37 @@ func (c *Controller) Run(ctx context.Context) {
 	<-ctx.Done()
 }
 
+// lookupBundleByDeploymentByIndex returns a function that can be used to perform lookups of Bundles that contain
+// Deployment objects that reference ConfigMap/Secret objects.
+func (c *Controller) lookupBundleByDeploymentByIndex(byIndex byIndexFunc, indexName string, indexKey indexKeyFunc) func(runtime.Object) ([]runtime.Object, error) {
```
ychen-atlassian: This could probably be a nice util function in ctrl because we do something very similar in basically everywhere we have a LookupHandler

pkg/controller/bundlec/controller.go
```diff
@@ -66,23 +77,50 @@ type Controller struct {
...
 // Prepare prepares the controller to be run.
-func (c *Controller) Prepare(crdInf cache.SharedIndexInformer, resourceInfs map[schema.GroupVersionKind]cache.SharedIndexInformer) {
+func (c *Controller) Prepare(crdInf cache.SharedIndexInformer, resourceInfs map[schema.GroupVersionKind]cache.SharedIndexInformer) error {
 	c.crdContext, c.crdContextCancel = context.WithCancel(context.Background())
 	crdInf.AddEventHandler(&crdEventHandler{
 		controller: c,
 		watchers:   make(map[string]watchState),
 	})
-
+	deploymentInf := resourceInfs[apps_v1.SchemeGroupVersion.WithKind("Deployment")]
```
nilebox: nit: I know that k/k monorepo doesn't have constants for resource kinds, but we could define our own constants and use them?

ash2k: Yes. A separate PR that updates all places with constants would be nice.

ash2k:
> @ash2k don't we need a similar fix for Secret -> ServiceInstance -> Bundle propagation?

@nilebox yes, good catch. We need the same fix there. https://github.com/atlassian/smith/issues/402

pkg/controller/bundlec/controller.go
ash2k: https://github.com/atlassian/smith/issues/403

ash2k:
I've added a test, ready for re-review. No code modifications were required. I could write a unit test but that would require a massive amount of setup code. I tried to re-use the `testCase` type that we have, but substituting workqueue with a mock one didn't work because it's copied to the lookup handler before I can replace it (without re-doing all the plumbing).

I'm going to do the same for `Secret`->`ServiceInstance` once this is merged.

pkg/controller/bundlec/controller.go
ash2k: I have checked the places where `LookupHandler` is used, but all of them are not like this one. They are simpler. We have duplication there and we should DRY it, but the logic is not the same as here. Or I'm missing something?

cmd/smith/app/bundle_controller.go
```diff
@@ -226,7 +227,10 @@ func (c *BundleControllerConstructor) New(config *ctrl.Config, cctx *ctrl.Contex
 		Broadcaster: broadcaster,
 		Recorder:    recorder,
 	}
-	cntrlr.Prepare(crdInf, resourceInfs)
+	err = cntrlr.Prepare(crdInf, resourceInfs)
```
brjacobsatlassian: :thinking: this should have been a warning before...

it/deployment_dependencies_test.go
```diff
@@ -0,0 +1,196 @@
...
+			APIVersion: smith_v1.BundleResourceGroupVersion,
+		},
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name: "bundle-dt",
+		},
+		Spec: smith_v1.BundleSpec{
+			Resources: []smith_v1.Resource{
+				{
+					Name: deploymentResourceName,
+					Spec: smith_v1.ResourceSpec{
```
brjacobsatlassian: We're going to have to make builders for this stuff or put it in files. I'm realizing there's.... a lot of boilerplate.

cmd/smith/app/bundle_controller.go
ash2k: It wasn't returning an error before this PR.

PR #408, opened 2019-01-24
Should check nil pointer for deletions

ychen-atlassian:


pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -246,7 +246,11 @@ func (st *bundleSyncTask) findObjectsToDelete() error {
 			gvk = res.Spec.Object.GetObjectKind().GroupVersionKind()
 			name = res.Spec.Object.(meta_v1.Object).GetName()
 		} else if res.Spec.Plugin != nil {
-			gvk = st.pluginContainers[res.Spec.Plugin.Name].Plugin.Describe().GVK
+			plugin, ok := st.pluginContainers[res.Spec.Plugin.Name]
+			if !ok {
+				continue
```
nilebox: We should return some error?

nilebox: ok, I guess the comment below explains it
```
			// neither "object" nor "plugin" field is specified. This shouldn't really happen (schema), but we
			// ignore the error and continue collecting objects. Even if not caught by the schema, this error
			// must have been reported earlier while processing this resource.
continue
```
copy this comment?

pkg/controller/bundlec/bundle_sync_task.go
```diff
@@ -246,7 +246,11 @@ func (st *bundleSyncTask) findObjectsToDelete() error {
 			gvk = res.Spec.Object.GetObjectKind().GroupVersionKind()
 			name = res.Spec.Object.(meta_v1.Object).GetName()
 		} else if res.Spec.Plugin != nil {
-			gvk = st.pluginContainers[res.Spec.Plugin.Name].Plugin.Describe().GVK
+			plugin, ok := st.pluginContainers[res.Spec.Plugin.Name]
+			if !ok {
+				return errors.Errorf("plugin %q is not a valid plugin", res.Spec.Plugin.Name)
```
ash2k: This should not break the loop - see below. This error should be ignored because it must have been reported earlier.

nilebox: Need a similar comment as below then, justifying a silent `continue`

ash2k: I was not right, `continue` is not the right behaviour here. This loops filters out things that should not be deleted. So if we continue, we might delete a thing that should not be deleted. Thanks @yiyangc91 for catching this.

PR #436, opened 2019-06-27
Update stuff

ash2k:


.travis.yml
```diff
@@ -56,28 +60,24 @@ script:
...
+  - helm install svc-cat/ups-broker --name ups-broker --namespace ups-broker
   # build container and run checks while service catalog and UPS are starting
   - make docker
   - make check
-  - |
-    kubectl create -f it/sc/ups-clusterservicebroker.yaml
-    sleep 5
-    make integration-test-sc-ci
+  - kubectl create -f it/sc/ups-clusterservicebroker.yaml
+  - sleep 5
```
nilebox: there should be more reliable way of checking that broker has been processed - check it's status?

